PR #3, opened 2015-05-04
Rename *Result messages

esteve:
@dirk-thomas @tfoote @wjwwood


dirk-thomas:
Should the `SetParametersResult` SetParametersResult be names singular (`SetParameterResult`) since it only describes the result of a single set parameter? The service then uses a list of these.

wjwwood:
It depends on how you interpret it, because `SetParametersResult` should read as "a `Result` of `SetParameters`", but if you change it to `SetParameterResult` then it would be "a `Result` of `SetParameter`", and `SetParameter` is not a service that exists. However, you could consider `SetParameterResult` to me "a `Result` of trying to set a parameter", where `SetParameter` is not a proper thing, but a reference to an attempted action.

I'd vote to leave it as is, because one Service uses a list of them, but the `SetParametersAtomically` uses only one, yet it describes setting multiple services, not just one.

dirk-thomas:
My interpretation was "the result of setting a parameter". I found it unnatural that the "result for set parameters" contains only the result of setting a single parameter.

tfoote:
Without the full phrase it's going to be slightly ambiguous anyway. I don't think there's a "correct" solution. 

+1

PR #13, opened 2016-12-13
use uint8 instead of deprecated byte

mikaelarguedas:
don't use deprecated byte but uint8

connects to https://github.com/ros2/rosidl/issues/190

### DIFF
--- a/rcl_interfaces/msg/ParameterValue.msg
+++ b/rcl_interfaces/msg/ParameterValue.msg
@@ -10,4 +10,4 @@ bool bool_value
 int64 integer_value
 float64 double_value
 string string_value
-byte[] bytes_value
+uint8[] bytes_value


mikaelarguedas:
I kept it as is so that people feel like they can pass hexadecimal values without worrying about sign.
This message being used only on the ROS2 side so we could change the field name without resulting in changes in the bridge but it seemed to be a less intrusive change (bytes are already treated as uint8 in the ROS2 code using parameters so it is transparent)

mikaelarguedas:
this also relates to the [upcoming C parameter](https://github.com/ros2/rcl/pull/54) implementation that may require this change and / or additional changes to the possible parameter types (e.g. arrays of parameters)

clalancette:
We haven't done this until now, so I'm going to venture to say we aren't going to do it.  Because of that, I'm going to close this out, but feel free to reopen if you are going to pursue it in the future.

PR #16, opened 2017-04-03
adding test_msgs to rcl_interfaces repo

mikaelarguedas:
replaces https://github.com/ros2/common_interfaces/pull/30
Fix ros2/system_tests#191
Connects to ros2/system_tests#191


dirk-thomas:
Once the message definitions and fixtures have been move https://github.com/ros2/ros2cli/blob/9fce6a08381bb67cf17e7160b0b2c6556ffec1e6/ros2topic/package.xml#L16 should switch to the new location.

dirk-thomas:
I think it would be valuable to preserve the history of the moved content. These files have quite some history which would be valuable for future debugging / tracing.

mikaelarguedas:
history of the files has been propagated, ready for review

test_msgs/package.xml
```diff
@@ -0,0 +1,24 @@
+<?xml version="1.0"?>
+<?xml-model href="http://download.ros.org/schema/package_format2.xsd" schematypens="http://www.w3.org/2001/XMLSchema"?>
+<package format="2">
+  <name>test_msgs</name>
+  <version>0.0.0</version>
+  <description>A package containing message definitions and fixtures used exclusively for testing purposes.</description>
+  <maintainer email="karsten@osrfoundation.org">Karsten Knese</maintainer>
```
dirk-thomas, line 7: Not sure why you picked @Karsten1987 but I don't mind :wink:

test_msgs/CMakeLists.txt
```diff
@@ -0,0 +1,53 @@
...
+
+rosidl_generate_interfaces(${PROJECT_NAME}
+  ${msg_files}
+  ${srv_files}
+  DEPENDENCIES builtin_interfaces
+  ADD_LINTER_TESTS
+)
+
+install(DIRECTORY src/test_msgs
+  DESTINATION "${PYTHON_INSTALL_DIR}/")
```
dirk-thomas, line 46: This Python package doesn't have an `__init__.py` file. Should it?

dirk-thomas (approved):

test_msgs/CMakeLists.txt
dirk-thomas: Nvm, just found the file being added in a later commit.

test_msgs/package.xml
mikaelarguedas: he is the one that started the original work in https://github.com/ros2/common_interfaces/pull/30 and I took over, I figured he'd like to keep maintainership :stuck_out_tongue_winking_eye:

dirk-thomas: > I figured he'd like to keep maintainership

Very generous of you :laughing: Works for me...

PR #18, opened 2017-08-16
Provide defaults for ParameterValue fields

dhood:
Even though only one field for a particular parameter is intended to be used at a time, to prevent serialisation of uninitialised fields we put defaults for all of them.

This is an alternative to https://github.com/ros2/rclcpp/pull/358, both designed to fix the valgrind error referenced in that PR.

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/3026/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/437/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/2434/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/3108/)

### DIFF
--- a/rcl_interfaces/msg/ParameterValue.msg
+++ b/rcl_interfaces/msg/ParameterValue.msg
@@ -6,7 +6,7 @@
 uint8 type
 
 # "Variant" style storage of the parameter value.
-bool bool_value False
+bool bool_value false
 int64 integer_value 0
 float64 double_value 0.0
 string string_value ""
--- a/rcl_interfaces/msg/ParameterValue.msg
+++ b/rcl_interfaces/msg/ParameterValue.msg
@@ -6,8 +6,8 @@
 uint8 type
 
 # "Variant" style storage of the parameter value.
-bool bool_value
-int64 integer_value
-float64 double_value
-string string_value
-byte[] bytes_value
+bool bool_value False
+int64 integer_value 0
+float64 double_value 0.0
+string string_value ""
+byte[] bytes_value []
--- a/rcl_interfaces/msg/ParameterValue.msg
+++ b/rcl_interfaces/msg/ParameterValue.msg
@@ -10,4 +10,4 @@ bool bool_value False
 int64 integer_value 0
 float64 double_value 0.0
 string string_value ""
-byte[] bytes_value []
+byte[] bytes_value


rcl_interfaces/msg/ParameterValue.msg
```diff
@@ -6,8 +6,8 @@
...
-bool bool_value
-int64 integer_value
-float64 double_value
-string string_value
-byte[] bytes_value
+bool bool_value False
+int64 integer_value 0
+float64 double_value 0.0
+string string_value ""
+byte[] bytes_value []
```
dirk-thomas: The default value for the array is unnecessary. An array is always empty by default.

rcl_interfaces/msg/ParameterValue.msg
```diff
@@ -6,8 +6,8 @@
 uint8 type
 
 # "Variant" style storage of the parameter value.
-bool bool_value
-int64 integer_value
-float64 double_value
-string string_value
+bool bool_value False
```
mikaelarguedas: Please use `false`(lowercase) to match the [design specification](http://design.ros2.org/articles/interface_definition.html)

dhood: to clarify, is your comment just for style? or is `False` not supposed to work? (it seems to work)

mikaelarguedas: It is not supposed to work given that our specification defines `true`, `false`, `1` and `0` as the only allowed valued for `bool`. It happens to work because we account for user error in our parser by [lowercasing](https://github.com/ros2/rosidl/blob/af7e2700ab30a28b6d9867508ae74887fa45233d/rosidl_parser/rosidl_parser/__init__.py#L496-L502) the value. But it is an implementation detail and would not work for any other parser written following the spec so I think we should not use it in our messages

dhood: ok thanks for the explanation -- just wanted to check we know why it's working as is. fixed in https://github.com/ros2/rcl_interfaces/pull/18/commits/43081b34c5e102c2575861c02c090e46044b9c5c

mikaelarguedas (approved):
lgtm

dirk-thomas (approved):

rcl_interfaces/msg/ParameterValue.msg
```diff
@@ -6,8 +6,8 @@
 uint8 type
```
dirk-thomas, line 6: Actually should we initialize this too?

mikaelarguedas: yeah good point, I guess that PARAMETER_NOT_SET would be a good default. Not sure how we refer to message constants from another msg file though.

dirk-thomas: Hm, I guess we can't at the moment...

dhood: My reasoning for not putting a default for this was that, while all-but-one of the `_value` fields will be unused, the `type` is not "optional" so the client library should always initialise it itself

mikaelarguedas: > Hm, I guess we can't at the moment...

Should we open a ticket for this somewhere (rosidl?) that would be a useful feature

dhood: I've got two thumbs up on my last comment so will consider this question resolved. (in the future, reactions don't send notifications (at least not to me!) so comments would be appreciated)

mikaelarguedas: :+1: sorry forgot about that, yeah it would be great if github allowed to enable notification for "reactions"

dhood:
I was running some tests on other PRs and wanted to know that this wasn't on master for the CI. I'll merge this now.

Since @mikaelarguedas pointed out that the default values we specify here are the same as the defaults for those message types in general, I've opened https://github.com/ros2/rcl_interfaces/issues/19 to track that this can be reverted if/when default-initialised messages can be created in client libraries

PR #23, opened 2017-10-31
[rcl_interfaces] Homogenize iface definition

mikaelarguedas:
This is just cleanup
- first commit make msg/srv definition consistent https://github.com/ros2/rcl_interfaces/commit/27920fa153a5b1506a748576853bc696ebb72687
- second commit moves the repo's README.md to the right folder https://github.com/ros2/rcl_interfaces/commit/89da210a9770da7a3de8a49e496a50005b8eb093
- thiird commit creates a generic README for the entire repo https://github.com/ros2/rcl_interfaces/commit/5cd70c2557e367a68ec96059fbe9a2e848589804


README.md
```diff
@@ -1,31 +1,2 @@
...
-
- * `get_parameters`: `GetParameters`
-  * The service to get the value of parameters which are set on this node.
- * `has_parameters`: `HasParameters`
-  * Query this node if specific parameters are set.
- * `list_parameters`: `ListParameters`
-  * List the parameters on this node matching the filters.
- * `set_parameters`: `SetParameters`
-  * Set parameters on this node.
+A set of packages which contain interface files (.msg and .srv) used for the internal implementation and testing of the client libraries.
```
dirk-thomas: I wouldn't consider those "internal". Most of them define public interfaces, no?

mikaelarguedas: "used for the implementation and testing" then ?

dirk-thomas: "defining the interface across client libraries"?

mikaelarguedas: Hmm not sure because as you said it is user facing and currently most of them are not used to communicate across client libraries given that they're used mostly/(only?) in rclcpp.

@ros2/team maybe some native speakers have some insights ?

wjwwood: They're definitely public and I would say their primary use is implementation of rcl concepts rather than testing, though they could be used for that too I guess.

"This repository contains a set of packages which themselves contain interface files (i.e. `.msg` and `.srv` files) used to implement client library concepts like parameters."

mikaelarguedas: my goal here is to give a description of the content of the repository and not only describe the `rcl_interfaces` package, that's why I talked about testing (given that this repo also contains test_msgs).

I'm fine with "This repository contains a set of packages which themselves contain interface files (i.e. .msg and .srv files) used to implement client library concepts" I just wanted to clarify the scope

wjwwood: Oh, I was confused by the diff and I didn't know there were test messages here. In that case I agree, something like:

"This repository contains a set of packages which themselves contain interface files (i.e. `.msg` and `.srv` files) which are used both to implement client library concepts and in testing."

Or something like that.

tfoote: I've updated it with approximately the verbiage of above.

mikaelarguedas:
Thanks @tfoote 
@ros2/team reay for review

wjwwood (approved):

PR #25, opened 2017-11-27
Extend service test suite

mikaelarguedas:
We currently don't have any service tests for arrays or nested messages, this PR extends the service test suites to match the current message test suite coverage.

Leaving this in progress until we have `wait_for_service` in python as these tests have potential for flakiness (should be merged after ros2/system_tests#244)

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/3614/)

All platforms:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/3616/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/799/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/2946/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/3702/)


mikaelarguedas:
ros2/system_tests#244 has merged, placing this in review

dirk-thomas:
A lot of the service fixtures are very similar / copy-n-paste of some message fixtures. I would suggest to refactor the message fixtures to e.g. accept a template type in order to reuse all the code setting the member variables to some values.

mikaelarguedas:
> A lot of the service fixtures are very similar / copy-n-paste of some message fixtures. I would suggest to refactor the message fixtures to e.g. accept a template type in order to reuse all the code setting the member variables to some values.

@dirk-thomas I did this as it seems that we wanted to have different fixtures for request, replies and pub/sub. That's what was done in the [system_tests branch](https://github.com/ros2/system_tests/commit/e5f10280aa8a9880dc1be18c42a2d59dfdc8efd2) that this PR replaces, I guess the goal was to ensure to detect issues like receiving the content of the request rather than the one of the reply. If I template the functions on the message type, we will be constrained to use the same fixtures everywhere or need extra code to populate the ones that we want different (and for cases whre we want to send smaller, less messages in service tests than in pub/sub tests)

That being said if we want to use the same fixtures everywhere and write this extra logic in the (I guess very few) cases where we want to, I'm totally fine templating them all.

mikaelarguedas:
currently blocked by CI roundtrip time. Putting this back in backlog

hidmic:
@dirk-thomas looks like some of this PR could be resurrected and migrated to https://github.com/ros2/test_interface_files. Any objections?

dirk-thomas:
> looks like some of this PR could be resurrected and migrated to https://github.com/ros2/test_interface_files.

Absolutely, `test_interface_files` is the right place to add these. For each new interface file it should be clearly stated why it is needed since the goal is to keep the number of test interfaces as small as possible.

jacobperron:
FYI, I've added `Arrays.srv` in https://github.com/ros2/test_interface_files/pull/7 in relation to a bugfix.

hidmic:
> For each new interface file it should be clearly stated why it is needed since the goal is to keep the number of test interfaces as small as possible.

Well, this PR's intent seems to have been to introduce service equivalents for all existing test messages at the time. Based on that, I'd do the same on `test_interface_files`. It seems to be the case for the few services already there too. Does that sound like a reasonable extension to our test base @dirk-thomas ?

dirk-thomas:
> this PR's intent seems to have been to introduce service equivalents for all existing test messages at the time.

As mentioned before:

> For each new interface file it should be clearly stated why it is needed since the goal is to keep the number of test interfaces as small as possible.

Just adding them to by symmetric with service isn't a good enough reason imo.

hidmic:
@dirk-thomas Thus my question. **It will be simply symmetric**. If that's not a good enough reason i.e. we trust that those IDL types working for messages will also work for services with no additional testing, then we don't need to migrate anything.

dirk-thomas:
> It will be simply symmetric.

Ok, lets close this PR then.

Please feel free to comment with more information and we can consider reopening it.

PR #30, opened 2017-12-05
switch to send clock with type rcl_interfaces/Time

tfoote:
Connects to ros2/rclcpp#420

Datatype for the workaround.


dirk-thomas:
I propose to close this in favor of ros2/rmw_opensplice#217.

Issue #31, opened 2018-01-09
Include path order references underlay before overlay


ayrton04:
## Bug report

**Required Info:**

- Operating System:
  - Ubuntu 16.04
- Installation type:
  - Binaries with an overlay containing `rcl_interfaces`
- Version or commit hash:
  - `param-arrays` branch on github.com:ayrton04/rcl_interfaces.git

#### Steps to reproduce issue
```
mkdir -p ~/ament_test/src
cd ~/ament_test/src
git clone git@github.com:ayrton04/rcl_interfaces.git -b param-arrays
cd ..
source /opt/ros/ardent/setup.bash
ament build
```

#### Expected behavior
Successful build

#### Actual behavior
```
[ 83%] Building C object CMakeFiles/rcl_interfaces__rosidl_typesupport_introspection_c__pyext.dir/rosidl_generator_py/rcl_interfaces/msg/_parameter_value_s.c.o
/home/automatom/ros2/build/rcl_interfaces/rosidl_generator_py/rcl_interfaces/msg/_parameter_value_s.c: In function ‘rcl_interfaces_parameter_value__convert_from_py’:
/home/automatom/ros2/build/rcl_interfaces/rosidl_generator_py/rcl_interfaces/msg/_parameter_value_s.c:111:61: error: ‘rcl_interfaces__msg__ParameterValue {aka struct rcl_interfaces__msg__ParameterValue}’ has no member named ‘byte_array_value’
 if (!rosidl_generator_c__byte__Array__init(&(ros_message->byte_array_value), size))
```

#### Additional information
Please see [this](https://answers.ros.org/question/278910/ros2-ament-overlay-issue-with-messages/) ROS answers question for more details.

FYI @wjwwood

wjwwood:
From my answer...

This is what the cc line looks like on my machine:

```
/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/cc
-Drcl_interfaces__rosidl_typesupport_introspection_c__pyext_EXPORTS
-I/Users/william/ament_test/build/rcl_interfaces/rosidl_typesupport_c
-I/Users/william/ament_test/build/rcl_interfaces/rosidl_typesupport_introspection_c
-I/Users/william/ros2_ws/install/include  <-- this has the old header
-I/Users/william/ament_test/build/rcl_interfaces/rosidl_generator_c  <-- this has the new header
[...]

PR #32, opened 2018-01-09
add array parameter types as defined in https://github.com/ros2/design/pull/152

mikaelarguedas:
This is to support @ayrton04's work on implementing parameter arrays


ayrton04:
Do we also need to update `ParameterValue.msg`, or is it only meant do be defining primitive values?

https://github.com/ros2/rcl_interfaces/compare/master...ayrton04:param-arrays?expand=1

mikaelarguedas:
> Do we also need to update `ParameterValue.msg`, or is it only meant do be defining primitive values?

Absolutely :+1: updated accordingly

ayrton04 (approved):
LGTM (or, at least, it matches what I have locally).

mikaelarguedas:
@ayrton04 I split and merged the renaming of the existing parameter array. I'll leave this one open and we will merge it along the corresponding `rclcpp` PR you're working on

ayrton04:
Awesome! I’m blocked at the moment because of this:

https://github.com/ros2/rcl_interfaces/issues/31

mikaelarguedas:
Ack, that's indeed a problem that need to be solved.
Though we do recommend people making contributions to develop and test their features in a workspace tracking master and built from source as ardent code and API will be out of sync pretty quick.
If it helps there's you can setup an environment development using [this Dockerfile](https://github.com/osrf/docker_images/tree/master/ros2/source) (to track master you will need to replace `release-latest` with `master` [here](https://github.com/osrf/docker_images/blob/590a0ee1033a29b1a8aacda0030b0680a1e43e39/ros2/source/source/Dockerfile#L82))

ayrton04:
Noted. Thanks for the heads-up, and for the fix!

ayrton04:
Just a status update: I believe I'm all done with the rclcpp implementation, and am adding tests for `ParameterVariant`.

sloretz (approved):

PR #35, opened 2018-05-30
Add read_only parameter metadata

sloretz:
WIP until there is code in rclcpp supporting this. Adds a boolean indicating if a parameter is allowed to change after it has been initialized. The default is `true`. If `false` then the parameter can be set via yaml passed to the node, but it's value may not be changed for the life of the node.

connects to ros2/rclcpp#495

### DIFF
--- a/rcl_interfaces/msg/ParameterDescriptor.msg
+++ b/rcl_interfaces/msg/ParameterDescriptor.msg
@@ -2,5 +2,7 @@
 
 string name
 uint8 type # Enum defined in ParameterType.msg
+# If 'false' then the value cannot change after it has been initialized
+bool mutable_after_initialization true
 
 # TODO: add additional meta data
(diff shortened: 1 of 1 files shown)


rcl_interfaces/msg/ParameterDescriptor.msg
```diff
@@ -2,5 +2,7 @@
 
 string name
 uint8 type # Enum defined in ParameterType.msg
+# If 'false' then the value cannot change after it has been initialized
+bool mutable_after_initialization true
```
mikaelarguedas: Do you think that `read_only` could be a descriptive enough name ? (and defaulting to `false` if we change the name)

sloretz: I think so. I think it's possible someone might think `read_only` means a run-time user can't initialize it with a value, or they might think a run-time user can't change the value but the node itself might change it and report changes on the `parameter_events` topic. However, the documentation can explain that here.

`dynamic` defaulting to `true` is another option, but that term relies on associating it with `dynamic_reconfigure` in ROS 1.

sloretz: changed to `read_only` in dab17415e19351a117cce3c69af67540f93bb9ba

jacobperron (approved):

PR #37, opened 2018-07-27
Only install the Python messages if PYTHON_INSTALL_DIR is defined

esteve:
This avoids installing the messages if Python does not exist (e.g. when building for Android, see https://github.com/esteve/ros2_java/issues/44)


dirk-thomas:
Thanks for the patch.

PR #38, opened 2018-07-27
Allow nanosec < 0

sloretz:
While fixing an issue found in ros2/rclcpp#525 I noticed the `nanosec` is unsigned. I'm not sure what this is supposed to mean. Does `nanosec` always have the same sign as `sec`, or is it always positive?

Code in rclcpp just adds the two together. However, ros2/rcutils#79 was opened to prevent mixing of signed/unsigned types. This PR makes `nanosec` signed, and adds a comment saying the two fields are to be combined by addition.

CI on ros2/rclcpp#527

### DIFF
--- a/builtin_interfaces/msg/Duration.msg
+++ b/builtin_interfaces/msg/Duration.msg
@@ -1,2 +1,4 @@
+# The length of a duration is sec and nanosec added together.
+# abs(length) can be less than abs(sec) if the sign of nanosec differs from sec.
 int32 sec
-uint32 nanosec
+int32 nanosec


dirk-thomas:
I think this will diverge from how DDS specifies `duration_t`. We should carefully consider the consequences,

dirk-thomas:
Imo it would be even better to keep it as unsigned and once `.idl` will allow us to specify that also constrain the upper limit to `<` 1 000 000 000.

wjwwood:
Changing it to signed isn't a good idea imo. You can't really compare it to the rcl ~~time point~~ duration because it's a single nanoseconds value, not part of a nanoseconds/seconds pair.

If we were to change this at all I'd advocate for a single nanoseconds signed 64-bit integer, but that would probably be too disruptive (different argument and out of scope for this issue at least).

~~I don't think we should compare to how DDS does it, since they're probably going to change their seconds component to unsigned to avoid the 2038 problem.~~ They won't change anything for 2038 with respect to durations, only time points.

Either way I wouldn't make the change proposed here.

wjwwood:
Err, I thought we were talking about a time point not a duration, so never mind the part about 2038 problem.

I still would leave this with an unsigned nanoseconds, though I could possibly be swayed if we could enforce the boundary on it that it's within `[0, 1e9)`.

dirk-thomas:
I think the comparison to DDS is valuable. Not necessarily as in "follow  what they do" but "consider why they chose what they did. The upper bound ensures that a duration message is automatically "nomalized" which I think is a good thing.

wjwwood:
> The upper bound ensures that a duration message is automatically "nomalized" which I think is a good thing.

Well, signed or unsigned doesn't change the boundary issue. It just means you need two boundaries or just one. Whereas having signed nanoseconds avoids mixing unsigned and signed arithmetic.

tfoote:
The other thing to consider for consistency is the same semantic as ROS1. If we're going to break that consistency there's not much reason to keep it in two fields. The 64bit signed nanoseconds is much simpler to process. But for now I'd suggest keeping it as is.

sloretz:
Ignoring the signed/unsigned mixing for the moment, if the duration message is signed `sec` and unsigned `nanosec`, then what is the length of time indicated by `.sec = -1, .nanosec=500000000`?

dirk-thomas:
> what is the length of time indicated by .sec = -1, .nanosec=500000000

`- 500 000 000 ns`

sloretz:
Thanks for the explanation.

PR #44, opened 2018-10-25
Msg/Services in action folder

sloretz:
This PR adds messages and services in an action folder

connects to ros2/rosidl#301


test_msgs/action/FibonacciGoalRequest.srv
```diff
@@ -0,0 +1,9 @@
+# Fields always added to goal requests
+uint8[16] uuid
```
jacobperron: What about using `action_msgs/GoalInfo` instead?  Although there isn't a planned use-case for timestamps in goal requests, it might be handy for debugging or new features.
ROS 1 goal requests also came with timestamps, and it looks like they were used for "bumping" goal handle destruction time:

https://github.com/ros/actionlib/blob/d317e63ebabc6c437f8b9f73ece06acf348700ea/include/actionlib/server/action_server_base.h#L226-L230

...I'm not sure if we want to bother with that logic though.

test_msgs/action/FibonacciGoalResult.srv
```diff
@@ -0,0 +1,6 @@
+# service respose fields are the same for all actions
```
jacobperron: "respose" -> "result request"

test_msgs/action/FibonacciGoalRequest.srv
```diff
@@ -0,0 +1,9 @@
...
+# Fields added by user code
+#goal definition
+int32 order
+---
+# service respose fields are the same for all actions
```
jacobperron: "response"

test_msgs/action/FibonacciGoalResult.srv
```diff
@@ -0,0 +1,6 @@
...
+action_msgs/GoalInfo goal
+---
+# All fields in response are user generated
```
jacobperron: I guess `action_msgs/GoalInfo` should be added to the response so it can be associated with the request by the client.

jacobperron (commented):
Some minor comments. The rest LGTM.

test_msgs/action/FibonacciGoalRequest.srv
sloretz: I'm not sure what that bumping logic does, but the overhead of using GoalInfo is very small so I'm fine with reusing it here

test_msgs/action/FibonacciGoalResult.srv
sloretz: I was assuming the goal result service would rely on the service API to associate the request and response. Services have a sequence number, but you might be right that it's easier if the GoalInfo is in the result too.

test_msgs/action/FibonacciGoalResult.srv
sloretz: 793d9d5eeba931b8de58435fb66ba348c78fd3f6

test_msgs/action/FibonacciGoalRequest.srv
sloretz: bd056c62be6d26d11b169b0559801a529035d113

test_msgs/action/FibonacciGoalResult.srv
sloretz: 41d60e526865f03f4be8d93f18eb762022e45141

test_msgs/action/FibonacciGoalRequest.srv
sloretz: 41d60e526865f03f4be8d93f18eb762022e45141

test_msgs/include/test_msgs/action/fibonacci.h
```diff
@@ -0,0 +1,73 @@
...
+
+
+#define ROSIDL_TYPESUPPORT_INTERFACE__TEST_MSGS__ACTION__FIBONACCI__GOAL_REQUEST_SYMBOL_NAME \
+  ROSIDL_TYPESUPPORT_INTERFACE__SERVICE_SYMBOL_NAME(rosidl_typesupport_c, test_msgs, action, FibonacciGoalRequest)
+
+#define ROSIDL_TYPESUPPORT_INTERFACE__TEST_MSGS__ACTION__FIBONACCI__GOAL_RESPONSE_SYMBOL_NAME \
+  ROSIDL_TYPESUPPORT_INTERFACE__SERVICE_SYMBOL_NAME(rosidl_typesupport_c, test_msgs, action, FibonacciGoalResponse)
+
+#define ROSIDL_TYPESUPPORT_INTERFACE__TEST_MSGS__ACTION__FIBONACCI__FEEDBACK_SYMBOL_NAME \
+  ROSIDL_TYPESUPPORT_INTERFACE__MESSAGE_SYMBOL_NAME(rosidl_typesupport_c, test_msgs, action, FibonacciGoalResponse)
```
hidmic: @sloretz `FibonacciGoalResponse` -> `FibonacciFeedback`?

sloretz: Fixed in 80647a87e30ffe2ad8cfcb9a43c1eb66d4c1d934

jacobperron (approved):
Pending CI

sloretz:
Closing. This was useful for testing the PRs undier ros2/rosidl#301, but doesn't serve a purpose now that  #47 is up

PR #45, opened 2018-10-26
Add message fixture for DynamicArrayStaticArrayPrimitivesNested

jacobperron:
Connects to 
This resolves CI failures introduced by #43.

CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/5361/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/2088/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/4451/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/5321/)

PR #52, opened 2018-11-15
add new test messages for arrays of arrays

Karsten1987:
We'd like to have two more test messages in the form of containers of containers. We would leverage them in our rosbag tests.

see https://github.com/ros2/rosbag2/pull/56#issuecomment-439093754
cc @anhosi

### DIFF
--- a/test_msgs/CMakeLists.txt
+++ b/test_msgs/CMakeLists.txt
@@ -21,11 +21,10 @@ find_package(action_msgs REQUIRED)
 find_package(rosidl_default_generators REQUIRED)
 
 set(msg_files
-  "msg/BoundedArrayBoundedArrayPrimitivesNested.msg"
+  "msg/BoundedArrayPrimitivesNested.msg"
   "msg/BoundedArrayNested.msg"
   "msg/BoundedArrayPrimitives.msg"
   "msg/Builtins.msg"
-  "msg/DynamicArrayDynamicArrayPrimitivesNested.msg"
   "msg/DynamicArrayNested.msg"
   "msg/DynamicArrayPrimitives.msg"
   "msg/DynamicArrayPrimitivesNested.msg"
@@ -35,6 +34,7 @@ set(msg_files
   "msg/Primitives.msg"
   "msg/StaticArrayNested.msg"
   "msg/StaticArrayPrimitives.msg"
+  "msg/StaticArrayPrimitivesNested.msg"
 )
 
 set(srv_files
--- a/test_msgs/CMakeLists.txt
+++ b/test_msgs/CMakeLists.txt
@@ -21,9 +21,11 @@ find_package(action_msgs REQUIRED)
 find_package(rosidl_default_generators REQUIRED)
 
 set(msg_files
+  "msg/BoundedArrayBoundedArrayPrimitivesNested.msg"
   "msg/BoundedArrayNested.msg"
   "msg/BoundedArrayPrimitives.msg"
   "msg/Builtins.msg"
+  "msg/DynamicArrayDynamicArrayPrimitivesNested.msg"
   "msg/DynamicArrayNested.msg"
   "msg/DynamicArrayPrimitives.msg"
   "msg/DynamicArrayPrimitivesNested.msg"
--- a/test_msgs/msg/DynamicArrayDynamicArrayPrimitivesNested.msg
+++ b/test_msgs/msg/DynamicArrayDynamicArrayPrimitivesNested.msg
@@ -1,1 +1,0 @@
-DynamicArrayPrimitives[] dynamic_array_dynamic_array_primitive_values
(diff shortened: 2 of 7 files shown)


test_msgs/msg/DynamicArrayDynamicArrayPrimitivesNested.msg
```diff
@@ -0,0 +1 @@
+DynamicArrayPrimitives[] dynamic_array_dynamic_array_primitive_values
```
jacobperron: This case already exists, equivalent to `DynamicArrayPrimitivesNested`:

https://github.com/ros2/rcl_interfaces/blob/2a85883cf64d9dfbe9f15a43966bdb00819aac58/test_msgs/msg/DynamicArrayPrimitivesNested.msg#L1

test_msgs/CMakeLists.txt
```diff
@@ -21,9 +21,11 @@ find_package(action_msgs REQUIRED)
 find_package(rosidl_default_generators REQUIRED)
 
 set(msg_files
+  "msg/BoundedArrayBoundedArrayPrimitivesNested.msg"
```
jacobperron: For consistency, I would name this `BoundedArrayPrimitivesNested`.

jacobperron (changes requested):
There are some consequences to adding new messages to this package:

1. It will cause [test_communications](https://github.com/ros2/system_tests/tree/master/test_communication) to fail, unless a few things are updated.
See my tickets related to previous tribulations:

- #45 
- ros2/system_tests#302

2. The build time increases significantly for all builds (exponential?) because of testing done in `test_communications`.
(@clalancette Correct me if I'm wrong).

Karsten1987:
Comments make sense to me, but I'll let @anhosi respond. 

If this doesn't get merged because of runtime reasons, we might have our own `rosbag2_test_msgs` paket. But i'd prefer not to.

jacobperron:
IMO, it makes sense to add the missing cases to this package for completeness:

- BoundedArrayPrimitivesNested
- StaticArrayPrimitivesNested

test_msgs/msg/DynamicArrayDynamicArrayPrimitivesNested.msg
anhosi: You are right, we missed this one.

test_msgs/CMakeLists.txt
anhosi: Makes sense, our suggestion was motivated by `DynamicArrayStaticArrayPrimitivesNested.msg`. But I would prefer your shorter suggestion.

Karsten1987:
I just updated the message files. Is there anything else to do with it in terms of test_communication?

jacobperron:
> I just updated the message files. Is there anything else to do with it in terms of test_communication?

@Karsten1987 
Support should be added for the new message types to the utility functions in this package:
- [message_fixtures.py](https://github.com/ros2/rcl_interfaces/blob/c48d2731548b36b1e8458e37f4f84d187ed8ed65/test_msgs/src/test_msgs/message_fixtures.py#L323)
- [message_fixtures.hpp](https://github.com/ros2/rcl_interfaces/blob/master/test_msgs/include/test_msgs/message_fixtures.hpp)

And then the publisher/subscriber tests in test_communications should be updated:

- https://github.com/ros2/system_tests/blob/5267cdd2ff2bb40adcee02e28f2c13901e51b6a2/test_communication/test/test_subscriber.cpp#L105
- https://github.com/ros2/system_tests/blob/5267cdd2ff2bb40adcee02e28f2c13901e51b6a2/test_communication/test/test_publisher_subscriber.cpp#L135
- https://github.com/ros2/system_tests/blob/5267cdd2ff2bb40adcee02e28f2c13901e51b6a2/test_communication/test/test_publisher.cpp#L73

An alternative to all of these changes is to let the publisher/subscriber tests in test_communications pass if the message type is unknown. Then any future additions to `test_msgs` wouldn't break integration tests. I.e. Return 0 [here](https://github.com/ros2/system_tests/blob/5267cdd2ff2bb40adcee02e28f2c13901e51b6a2/test_communication/test/test_subscriber.cpp#L144) and the same for the other test I've linked.

Karsten1987:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/5586/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/2220/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/4595/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/5465/)

jacobperron (approved):

Karsten1987:
new round of CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/5589/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/2223/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/4598/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/5468/)

Issue #50, opened 2018-11-14
rosgraph_msgs Log message type. 


mlautman:
I am working on a port of [rqt_console](https://github.com/ros-visualization/rqt_console) with @dirk-thomas which uses the `rosgraph_msgs/Log` message. I see that `rosgraph_msg/Log` hasn't been ported to ros2 yet and I was wondering if I should implement the Log msg or if there was a specific reason for not porting Log in rosgraph_msgs. As this is more of a design decision, I am happy to move the conversation to discourse if you believe that this should be a more general discussion.

tfoote:
I don't believe that there's any reason that it hasn't been ported except that it hasn't been needed yet. 

I would suggest coordinating with @nburek

He's working on several PRs right now to start publishing console outputs: 
https://github.com/ros2/rclcpp/pull/582
https://github.com/ros2/rcl/pull/327
https://github.com/ros2/rcutils/pull/127

nburek:
Hey Mike,

I am currently working on adding support for logging to a "rosout" topic and as part of that have ported over the Log message from rosgraph_msg. The message I'm working with looks mostly the same except that I've adjusted the level constants to match the enum values from the rcl code and I've removed the list of topics in the message as it didn't make sense to me that you would want that list with every log message. 

I was also looking at moving the message from "rosgraph_msg" into "common_interfaces/diagnostic_msgs" as logs seemed better grouped as a diagnostics tool than a rosgraph tool. 

I'll start working on getting a pull request created for my Log message port (it may take a couple of days to go through my company's open source approval process before I can officially make the pull request). 

I've updated the discourse topic discussing ROS2 logging with this information: https://discourse.ros.org/t/ros2-logging/6469

dirk-thomas:
Since the message should be used by `rcl` please add the Log message to [rcl_interfaces](https://github.com/ros2/rcl_interfaces/tree/master/rcl_interfaces) package. That one is a dependency of `rcl` already. Putting the message in `common_interfaces` would add another dependency to `rcl` and require `common_interfaces` to be released before `rcl` which isn't desired.

nburek:
Ok, makes sense. I'll add it to rcl_interfaces instead.

tfoote:
For consistency I'd suggest keeping the same package + name too.

dirk-thomas:
> For consistency I'd suggest keeping the same package + name too.

[...]

PR #53, opened 2018-11-15
Adding a message type for Log messages that will be sent out over the…

nburek:
Adding a message type for Log messages that will be sent out over the rosout topic.

Issue for this request: https://github.com/ros2/rcl_interfaces/issues/50

### DIFF
--- a/rcl_interfaces/CMakeLists.txt
+++ b/rcl_interfaces/CMakeLists.txt
@@ -12,6 +12,7 @@ endif()
 
 find_package(ament_cmake REQUIRED)
 find_package(rosidl_default_generators REQUIRED)
+find_package(builtin_interfaces REQUIRED)
 
 # ament_export_dependencies(rcl_interfaces)
 
@@ -32,6 +33,7 @@ rosidl_generate_interfaces(${PROJECT_NAME}
   "srv/ListParameters.srv"
   "srv/SetParametersAtomically.srv"
   "srv/SetParameters.srv"
+  DEPENDENCIES builtin_interfaces
 )
 
 # this must happen before the invocation of ament_package()
--- a/rcl_interfaces/CMakeLists.txt
+++ b/rcl_interfaces/CMakeLists.txt
@@ -18,6 +18,7 @@ find_package(rosidl_default_generators REQUIRED)
 rosidl_generate_interfaces(${PROJECT_NAME}
   "msg/IntraProcessMessage.msg"
   "msg/ListParametersResult.msg"
+  "msg/Log.msg"
   "msg/ParameterDescriptor.msg"
   "msg/ParameterEventDescriptors.msg"
   "msg/ParameterEvent.msg"
--- a/rcl_interfaces/msg/Log.msg
+++ b/rcl_interfaces/msg/Log.msg
@@ -1,0 +1,19 @@
+##
+## Severity level constants
+##
+byte DEBUG=10 #debug level
+byte INFO=20  #general level
+byte WARN=30  #warning level
+byte ERROR=40 #error level
+byte FATAL=50 #fatal/critical level
+
+##
+## Fields
+##
+Header header
+byte level
+string name # name of the node
+string msg # message
+string file # file the message came from
+string function # function the message came from
+uint32 line # line the message came from
--- a/rcl_interfaces/msg/Log.msg
+++ b/rcl_interfaces/msg/Log.msg
@@ -10,6 +10,7 @@ byte FATAL=50 #fatal/critical level
 ##
 ## Fields
 ##
+builtin_interfaces/Time stamp
 uint8 level
 string name # the name representing the logger this message came from
 string msg # message
--- a/rcl_interfaces/msg/Log.msg
+++ b/rcl_interfaces/msg/Log.msg
@@ -11,8 +11,8 @@ byte FATAL=50 #fatal/critical level
 ## Fields
 ##
 Header header
-byte level
-string name # name of the node
+uint8 level
+string name # the name representing the logger this message came from
 string msg # message
 string file # file the message came from
 string function # function the message came from
--- a/rcl_interfaces/msg/Log.msg
+++ b/rcl_interfaces/msg/Log.msg
@@ -10,7 +10,6 @@ byte FATAL=50 #fatal/critical level
 ##
 ## Fields
 ##
-Header header
 uint8 level
 string name # the name representing the logger this message came from
 string msg # message
--- a/rcl_interfaces/package.xml
+++ b/rcl_interfaces/package.xml
@@ -13,6 +13,8 @@
   <buildtool_depend>ament_cmake</buildtool_depend>
   <buildtool_depend>rosidl_default_generators</buildtool_depend>
 
+  <depend>builtin_interfaces</depend>
+
   <exec_depend>rosidl_default_runtime</exec_depend>
 
   <test_depend>ament_lint_auto</test_depend>


tfoote:
For reference this is a migration of the rosgraph_msgs/Log in ROS1: http://docs.ros.org/melodic/api/rosgraph_msgs/html/msg/Log.html
```
##
## Severity level constants
##
byte DEBUG=1 #debug level
byte INFO=2  #general level
byte WARN=4  #warning level
byte ERROR=8 #error level
byte FATAL=16 #fatal/critical level
##
## Fields
##
Header header
byte level
string name # name of the node
string msg # message 
string file # file the message came from
string function # function the message came from
uint32 line # line the message came from
string[] topics # topic names that the node publishes
```


The primary changes are changing the constant values and dropping the topics. 
I agree that dropping the topics makes sense. 

The enum values have changed from a bitmask like approach to a sequential approach with padding. This is matching the values in rcutils: https://github.com/ros2/rcutils/blob/master/include/rcutils/logging.h#L156-L165
They were set in https://github.com/ros2/rcutils/pull/57 to follow python's conventions: https://github.com/ros2/rcutils/pull/57#pullrequestreview-69952305

Since we're using an enum and I don't believe that we're planning to support custom logging levels I'd suggest that we restore the old enum values for consistency adding 0 for UNSET.

dirk-thomas:
> name of the node

This comment is only partially true. The name might be a node name but it can also be any other user defined string.

> Since we're using an enum and I don't believe that we're planning to support custom logging levels I'd suggest that we restore the old enum values.

The logging system already does support arbitrary numbers in between. So the constants should match the implementation exactly. The type should be changed from `byte` to `uint8` though since the data is actually numeric with its implicit order and not an opaque octet.

> for consistency adding 0 for UNSET.

Since there is no semantic associated to `0` atm I think there shouldn't be a constant defined in the message for it.

nburek:
> > name of the node
> 
> This comment is only partially true. The name might be a node name but it can also be any other user defined string.

I'll update the comment to mention that the field is meant to represent the name of the logger used to log this message. The logger name won't exactly match the node name anyways


> > Since we're using an enum and I don't believe that we're planning to support custom logging levels I'd suggest that we restore the old enum values.
> 
> The logging system already does support arbitrary numbers in between. So the constants should match the implementation exactly. The type should be changed from `byte` to `uint8` though since the data is actually numeric with its implicit order and not an opaque octet.
> 
> >

I'll change the type from `byte` to `uint8`.

mlautman:
Don't forget to add msg/Log.msg to the rosidl_generate_interfaces call

mlautman:
also, you need to replace `Header` with `common_interfaces/Header`

mlautman:
Header -> std_msgs/Header

nburek:
Thanks for the feedback @mlautman. I've added the Log.msg to the rosidl_generate_interface call in CMake. I've also removed the header from the Log message for now since it would add a dependency on std_msgs from common_interfaces. In the conversation on issue https://github.com/ros2/rcl_interfaces/issues/50, there was a desire to not have any new dependencies added in front of rcl.

mlautman:
Seems strange to not be able to depend on a core library like std_msgs. The header field is really useful

dirk-thomas:
> Seems strange to not be able to depend on a core library like std_msgs.

That statement is taken out of context. It is undesired to add a dependency to the `common_interfaces` package into the `rcl` packages. Since the log message should continue to use a header field the `std_msgs` package (or maybe just that specific message) could be moved into the `rcl_interfaces` repo...

nburek:
It may make sense for the Header message definition to be moved into rcl_interfaces. If that's the case then the Log message could add it back in as a field. However, should that be done for this pull request or should we move forward with the Log message as defined in this request and then create a new issue for the follow up work of moving either `std_msgs` or `Header` so they are part of the core dependencies?

dirk-thomas:
The `Header` also contains a frame id which isn't needed / used here. Therefore I would suggest to add a timestamp field instead.

nburek:
> The `Header` also contains a frame id which isn't needed / used here. Therefore I would suggest to add a timestamp field instead.

That may beg the question of if frame id belongs inside of `Header` or not. Judging by the comment in the .msg file, the frame id pertains to transformation data. Does it make sense to remove the frame id to make `Header` apply more generally or should header potentially be moved/renamed so that it's more apparent the Header is meant to be used with Transform messages.

wjwwood:
> That may beg the question of if frame id belongs inside of Header or not. Judging by the comment in the .msg file, the frame id pertains to transformation data. Does it make sense to remove the frame id to make Header apply more generally or should header potentially be moved/renamed so that it's more apparent the Header is meant to be used with Transform messages.

I would say no... The header is used mostly in the sensor messages where data both has a creation time and a physical origin location, in which case the frame id makes sense.

I think it's just convenient for people to grab the header when they needed a timestamp and/or a sequence number (sequence number is no longer in ROS 2's header message). Personally, I think this is a case of "Header abuse".

tfoote:
The std_msgs/Header has significant meaning for physically connected data which is almost all data at the robotic developer level or robot application level.  What we're doing here is much more internal to the middleware. We could define a new datatype but we've already deprecated the sequence number in the header and if you remove the frame_id it's just a timestamp. So just using a raw timestamp is actually really all that there is to a std_msgs/Header if we don't want the frame_id.

There's probably a slightly better name for std_msgs/Header but since it's used throughout the ROS ecosystem in it's current name, changing it is much more of a hassle than it's worth.

rcl_interfaces/msg/Log.msg
```diff
@@ -0,0 +1,18 @@
...
+##
+byte DEBUG=10 #debug level
+byte INFO=20  #general level
+byte WARN=30  #warning level
+byte ERROR=40 #error level
+byte FATAL=50 #fatal/critical level
+
+##
+## Fields
+##
```
tfoote, line 12:
```suggestion
##
builtin_interfaces/Time stamp
```

nburek:
Alright, good to know. I'll go ahead and add timestamp as a Time type field

rcl_interfaces/CMakeLists.txt
```diff
@@ -18,6 +18,7 @@ find_package(rosidl_default_generators REQUIRED)
 rosidl_generate_interfaces(${PROJECT_NAME}
```
mlautman:
```suggestion
find_package(builtin_interfaces REQUIRED)

rosidl_generate_interfaces(${PROJECT_NAME}
```

mlautman:
Add `find_package(builtin_interfaces REQUIRED)` after `find_package(ament_cmake REQUIRED)` and `  DEPENDENCIES builtin_interfaces` at the end of `rosidl_generate_interfaces`

nburek:
> Add `find_package(builtin_interfaces REQUIRED)` after `find_package(ament_cmake REQUIRED)` and ` DEPENDENCIES builtin_interfaces` at the end of `rosidl_generate_interfaces`

Thanks Mike. I'm actually in the middle of doing that now and verifying it builds locally. Will update pull request in a minute.

tfoote:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/5668/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/2276/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/4686/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/5561/)

tfoote (approved):
This lgtm and it's passing CI.

clalancette (approved):

PR #56, opened 2018-12-18
adding services for getting and setting logger levels for a node

mlautman:
Adds msg and srv types for getting and setting logger levels.


tfoote:
This looks good as a direct port, but I'm wondering if we might not want to rethink using a service for this to use a parameter API instead. https://github.com/ros2/rclpy/pull/254#issuecomment-448421271

clalancette:
Given the current DDS overhead of creating "default" services/topics, I'm not sure that this is a great idea.  Also, I do like @tfoote's idea of using a parameter for this instead.  Given that, and [this comment](https://github.com/ros2/rclpy/pull/254#issuecomment-456540563), I'm thinking we aren't going to use this as-is.  So because of all of that, I'm closing this out.  Feel free to reopen if you think this is in error.

PR #62, opened 2019-02-13
Add action with nested messages

sloretz:
connects to ros2/rosidl#348
CI at ros2/rosidl_typesupport#44

This PR should be backported to crystal if ros2/system_tests#330 is

### DIFF
--- a/test_msgs/action/NestedMessage.action
+++ b/test_msgs/action/NestedMessage.action
@@ -1,0 +1,8 @@
+# goal definition
+Builtins nested_field_no_pkg
+---
+# result definition
+test_msgs/Primitives nested_field
+---
+# feedback
+builtin_interfaces/Time nested_different_pkg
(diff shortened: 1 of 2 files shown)


test_msgs/action/NestedMessage.action
```diff
@@ -0,0 +1,8 @@
+# goal definition
+Builtins nested_field_no_pkg
```
jacobperron, line 2: To better cover the possible scenarios, what about adding all three types, `nested_field_no_pkg`, `nested_field`, and `nested_different_pkg` to all parts of the action spec (Goal, Result, and Feedback)?

sloretz: Added all types in all parts in c8917a01e560a9fef5c26fca8d270163bd2c5ef1

jacobperron (approved):
LGTM with green CI

PR #64, opened 2019-02-15
update test fixtures for char type

dirk-thomas:
Connected to [ros2/rosidl#334](https://github.com/ros2/rosidl/pull/334).


hidmic (commented):
LGTM, though many of the test values have changed e.g. `0x7F != 255`, `0x11 != 11`, etc. But I can't seem to find anything relying on them so I think we're fine.

dirk-thomas:
> though many of the test values have changed

Just because the type and valid value range has changed.

hidmic (approved):

hidmic:
Looks like I don't have true powers here :(

PR #68, opened 2019-03-26
Added preallocate test messages

gonzodepedro:
Signed-off-by: Gonzalo de Pedro <gonzalo@depedro.com.ar>

Add test_messages required for memory preallocation testing

Connects to /ros2/rmw_connext#346 and /ros2/rmw_connext#345


jacobperron:
I'm curious why the existing test messages are not sufficient. What case(s) are not covered by the existing messages?

mjcarroll:
I think that the existing ones are sufficient, and that this may not have been needed.  I think it was for additional testing to make sure that allocations didn't happen on larger bounded messages.

Issue #71, opened 2019-04-17
add lower / upper boundary to parameter descriptor


dirk-thomas:
Dynamic reconfigure in ROS 1 allows to define a lower / upper bound for numeric parameters. That information enables more usable UI elements like sliders.

While the logic to check if a specific parameter can be arbitrary complex narrowing the valid numeric values with a lower / upper bound already covers many use cases. Therefore the [ParameterDescriptor](https://github.com/ros2/rcl_interfaces/blob/1189517a0217a2cb9f78e71c2959542429ad9fc0/rcl_interfaces/msg/ParameterDescriptor.msg) should be extended to add this information.

For integer parameters obviously integer boundaries would be fitting. For floating point number the boundaries should be floating point numbers though in order to e.g. express a range [0, pi].

Since any fields are part of the message independent of the actual parameter type adding multiple different boundaries for each type isn't appealing since it increases the message size.

Therefore I propose to add lower and upper bounds with a floating point type. It enables the precise range necessary for floating point parameters and for integer parameters the conventions could be to round to the closest integer.

I am looking for early feedback on this so that the implementation can start beginning of next week.

wjwwood:
Obviously we don't need to sort out or act on everything in this comment below this line, but I just wanted to brain dump so we might consider these things while working on this and perhaps avoid needing to make more painful API breaks in the future.

----

I think it would be worth having a step as well, so you could do things like `[0.0, 1.0, 0.1]` or `[0, 100, 2]` for even numbers only. Dynamic reconfigure supported this.

The step could be documented that it needs to be unsigned and a value of `0` could be used to disable it for floating point and integers (though for integers 1 would be the same as 0).

A step value complicates verification a bit (step needs to be `<= distance(min, max)`), but you already need to check for special cases like: `min == max`, or if you want to support things like `[1, -1, ...]` (inverted slider) or `[1, 1, ...]` (only one valid value), so checking that the step is a valid value is not much more of a burden. It does, however, lead to interesting test cases like `[0, 1, 2]`.

----

> Since any fields are part of the message independent of the actual parameter type adding multiple different boundaries for each type isn't appealing since it increases the message size.

I agree having one for int and float is wasteful, but I'm not sure how important that is because I believe  this descriptor is going to be sent infrequently (it is not part of the parameter event payload for example). And so I think we should consider simplicity and expressiveness as part of the trade-off too.

----

[...]

PR #72, opened 2019-04-19
Parameters support ranges

wjwwood:
Based on the recommendation from @dirk-thomas here: https://github.com/ros2/rcl_interfaces/issues/71#issuecomment-484613435

So that we can iterate on it more effectively. See also the discussion in that issue.

This is currently targeted at the `read_only_parameters` branch because that one adds the `read_only` constraint already, but I will re-target the pull request to `master` once that one is merged.

Connects to #71

### DIFF
--- a/rcl_interfaces/CMakeLists.txt
+++ b/rcl_interfaces/CMakeLists.txt
@@ -17,6 +17,8 @@ find_package(builtin_interfaces REQUIRED)
 # ament_export_dependencies(rcl_interfaces)
 
 rosidl_generate_interfaces(${PROJECT_NAME}
+  "msg/FloatingPointRange.msg"
+  "msg/IntegerRange.msg"
   "msg/IntraProcessMessage.msg"
   "msg/ListParametersResult.msg"
   "msg/Log.msg"
@@ -27,8 +29,6 @@ rosidl_generate_interfaces(${PROJECT_NAME}
   "msg/ParameterType.msg"
   "msg/ParameterValue.msg"
   "msg/SetParametersResult.msg"
-  "msg/FloatingPointRange.msg"
-  "msg/IntegerRange.msg"
   "srv/DescribeParameters.srv"
   "srv/GetParameters.srv"
   "srv/GetParameterTypes.srv"
--- a/rcl_interfaces/CMakeLists.txt
+++ b/rcl_interfaces/CMakeLists.txt
@@ -27,6 +27,8 @@ rosidl_generate_interfaces(${PROJECT_NAME}
   "msg/ParameterType.msg"
   "msg/ParameterValue.msg"
   "msg/SetParametersResult.msg"
+  "msg/FloatingPointRange.msg"
+  "msg/IntegerRange.msg"
   "srv/DescribeParameters.srv"
   "srv/GetParameters.srv"
   "srv/GetParameterTypes.srv"
--- a/rcl_interfaces/msg/FloatingPointRange.msg
+++ b/rcl_interfaces/msg/FloatingPointRange.msg
@@ -1,10 +1,10 @@
 # Represents bounds and a step value for a floating point typed parameter.
 
 # Start value for valid values, inclusive.
-double from_value
+float64 from_value
 
 # End value for valid values, inclusive.
-double to_value
+float64 to_value
 
 # Size of valid steps between the from and to bound.
 # Step is considered to be a magnitude, therefore negative values are treated the
@@ -21,4 +21,4 @@ double to_value
 # not a multiple of the step, then the "to" bound will always be a valid value,
 # e.g. if the range is defined as {from_value: 2.0, to_value: 5.0, step: 2.0} then
 # the valid values will be 2.0, 4.0, and 5.0.
-double step
+float64 step
--- a/rcl_interfaces/msg/FloatingPointRange.msg
+++ b/rcl_interfaces/msg/FloatingPointRange.msg
@@ -1,14 +1,24 @@
 # Represents bounds and a step value for a floating point typed parameter.
 
-# Lower bound for valid values, inclusive.
-double minimum_value
+# Start value for valid values, inclusive.
+double from_value
 
-# Upper bound for valid values, inclusive.
-double maximum_value
+# End value for valid values, inclusive.
+double to_value
 
-# Size of valid steps between the lower and upper bound.
-# Step should be positive, and a step of zero implies continuous values.
-# Step size should be less than the distance between the minimum and maximum
-# values, and ideally would be an even multiple of that distance, but that
-# is not required.
+# Size of valid steps between the from and to bound.
+# Step is considered to be a magnitude, therefore negative values are treated the
+# same as positive values, and a step value of zero implies a continuous range of
+# values.
+# Ideally, the step would be less than or equal to the distance between the
+# bounds, as well as an even multiple of the distance between the bounds, but
+# neither are required.
+# If the absolute value of the step is larger than or equal to the distance
+# between the two bounds, then the bounds will be the only valid values,
+# e.g. if the range is defined as {from_value: 1.0, to_value: 2.0, step: 5.0} then
+# the valid values will be 1.0 and 2.0.
+# If the step is less than the distance between the bounds, but the distance is
+# not a multiple of the step, then the "to" bound will always be a valid value,
+# e.g. if the range is defined as {from_value: 2.0, to_value: 5.0, step: 2.0} then
+# the valid values will be 2.0, 4.0, and 5.0.
 double step
--- a/rcl_interfaces/msg/FloatingPointRange.msg
+++ b/rcl_interfaces/msg/FloatingPointRange.msg
@@ -1,0 +1,14 @@
+# Represents bounds and a step value for a floating point typed parameter.
+
+# Lower bound for valid values, inclusive.
+double minimum_value
+
+# Upper bound for valid values, inclusive.
+double maximum_value
+
+# Size of valid steps between the lower and upper bound.
+# Step should be positive, and a step of zero implies continuous values.
+# Step size should be less than the distance between the minimum and maximum
+# values, and ideally would be an even multiple of that distance, but that
+# is not required.
+double step
--- a/rcl_interfaces/msg/IntegerRange.msg
+++ b/rcl_interfaces/msg/IntegerRange.msg
@@ -1,0 +1,15 @@
+# Represents bounds and a step value for an integer typed parameter.
+
+# Lower bound for valid values, inclusive.
+int64 minimum_value
+
+# Upper bound for valid values, inclusive.
+int64 maximum_value
+
+# Size of valid steps between the lower and upper bound.
+# Step should be positive, and a step of zero implies a step size of one, i.e.
+# all integer values between the bounds are valid.
+# Step size should be less than the distance between the minimum and maximum
+# values, and ideally would be an even multiple of that distance, but that
+# is not required.
+int64 step
--- a/rcl_interfaces/msg/IntegerRange.msg
+++ b/rcl_interfaces/msg/IntegerRange.msg
@@ -1,15 +1,22 @@
 # Represents bounds and a step value for an integer typed parameter.
 
-# Lower bound for valid values, inclusive.
-int64 minimum_value
+# Start value for valid values, inclusive.
+int64 from_value
 
-# Upper bound for valid values, inclusive.
-int64 maximum_value
+# End value for valid values, inclusive.
+int64 to_value
 
-# Size of valid steps between the lower and upper bound.
-# Step should be positive, and a step of zero implies a step size of one, i.e.
-# all integer values between the bounds are valid.
-# Step size should be less than the distance between the minimum and maximum
-# values, and ideally would be an even multiple of that distance, but that
-# is not required.
-int64 step
+# Size of valid steps between the from and to bound.
+# A step value of zero implies a continuous range of values.
+# Ideally, the step would be less than or equal to the distance between the
+# bounds, as well as an even multiple of the distance between the bounds, but
+# neither are required.
+# If the absolute value of the step is larger than or equal to the distance
+# between the two bounds, then the bounds will be the only valid values,
+# e.g. if the range is defined as {from_value: 1, to_value: 2, step: 5} then
+# the valid values will be 1 and 2.
+# If the step is less than the distance between the bounds, but the distance is
+# not a multiple of the step, then the "to" bound will always be a valid value,
+# e.g. if the range is defined as {from_value: 2, to_value: 5, step: 2} then
+# the valid values will be 2, 4, and 5.
+uint64 step
--- a/rcl_interfaces/msg/ParameterDescriptor.msg
+++ b/rcl_interfaces/msg/ParameterDescriptor.msg
@@ -1,8 +1,9 @@
 # This is the message to communicate a parameter's descriptor.
 
 string name
-uint8 type # Enum defined in ParameterType.msg
-# If 'true' then the value cannot change after it has been initialized
-bool read_only false
 
-# TODO: add additional meta data
+# Enum values are defined in the `ParameterType.msg` message.
+uint8 type
+
+# If 'true' then the value cannot change after it has been initialized.
+bool read_only false
--- a/rcl_interfaces/msg/ParameterDescriptor.msg
+++ b/rcl_interfaces/msg/ParameterDescriptor.msg
@@ -30,5 +30,3 @@ FloatingPointRange[<=1] floating_point_range
 # IntegerRange consists of a from_value, a to_value, and a step.
 # FloatingPointRange and IntegerRange are mutually exclusive.
 IntegerRange[<=1] integer_range
-
-# more constraints to follow in the future...
--- a/rcl_interfaces/msg/ParameterDescriptor.msg
+++ b/rcl_interfaces/msg/ParameterDescriptor.msg
@@ -5,5 +5,15 @@ string name
 # Enum values are defined in the `ParameterType.msg` message.
 uint8 type
 
+# Parameter constraints
+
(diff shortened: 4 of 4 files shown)


wjwwood:
Proposed a change to add description strings here: https://github.com/ros2/rcl_interfaces/pull/73

rcl_interfaces/msg/IntegerRange.msg
```diff
@@ -0,0 +1,15 @@
...
+# Upper bound for valid values, inclusive.
+int64 maximum_value
+
+# Size of valid steps between the lower and upper bound.
+# Step should be positive, and a step of zero implies a step size of one, i.e.
+# all integer values between the bounds are valid.
+# Step size should be less than the distance between the minimum and maximum
+# values, and ideally would be an even multiple of that distance, but that
+# is not required.
+int64 step
```
wjwwood: Should this be unsigned?

rcl_interfaces/msg/ParameterDescriptor.msg
```diff
@@ -1,8 +1,19 @@
 # This is the message to communicate a parameter's descriptor.
 
 string name
-uint8 type # Enum defined in ParameterType.msg
-# If 'true' then the value cannot change after it has been initialized
+
+# Enum values are defined in the `ParameterType.msg` message.
+uint8 type
```
wjwwood, line 6: Anyone know why we decided to not place a `type` or `data` field in `ParameterType.msg` and then use `ParameterType type` here? Just to avoid `...type.data` or `...type.type`?

It would be nice to not have to imply where the constants for this field are using a comment.

rcl_interfaces/msg/ParameterDescriptor.msg
```diff
@@ -1,8 +1,19 @@
...
+
+# Parameter constraints
+
+# If 'true' then the value cannot change after it has been initialized.
 bool read_only false
 
-# TODO: add additional meta data
+# docs...
```
wjwwood: Maybe these don't need docs here? Since they have docs in their own message files.

rcl_interfaces/msg/FloatingPointRange.msg
```diff
@@ -0,0 +1,14 @@
+# Represents bounds and a step value for a floating point typed parameter.
+
+# Lower bound for valid values, inclusive.
+double minimum_value
+
+# Upper bound for valid values, inclusive.
+double maximum_value
```
wjwwood: Should we allow minimum value to be greater than and/or equal to maximum value?

dirk-thomas: I wouldn't constraint that. I larger minimum value could be presented in a UI as a slider with decrementing values from left to right.

rcl_interfaces/msg/IntegerRange.msg
dirk-thomas: I would keep it signed for the same reason as above.

I would suggest a default value of 1 though.

rcl_interfaces/msg/IntegerRange.msg
```diff
@@ -0,0 +1,15 @@
...
+
+# Lower bound for valid values, inclusive.
+int64 minimum_value
+
+# Upper bound for valid values, inclusive.
+int64 maximum_value
+
+# Size of valid steps between the lower and upper bound.
+# Step should be positive, and a step of zero implies a step size of one, i.e.
+# all integer values between the bounds are valid.
```
dirk-thomas: I would even leave out this restriction. Also at the moment it is ambiguous. If `min` is `-5` and `max` is `5` it reads like step can only be in between these values though choosing e.g. `8` shouldn't be prohibited.

If someone wants to specify a step larger than the interval why not let them?

rcl_interfaces/msg/ParameterDescriptor.msg
dirk-thomas: A brief description should be added since it can be represented as the doc string of that member in the generated code (which the comment in the submessage won't provide). It can also clarify why this is a bounded sequence.

rcl_interfaces/msg/IntegerRange.msg
wjwwood: Because, I don't even know how you'd represent that on a slider? The step is essentially capped at the distance between maximum and minimum, anything larger is effectively the same thing I guess, but it seems like something that is more likely to be an error than intentional.

rcl_interfaces/msg/IntegerRange.msg
wjwwood: What's the reason above? What is the meaning of a negative step value?

rcl_interfaces/msg/FloatingPointRange.msg
tfoote: I would advocate for not supporting inverting min and max values. If I told you a min values of 10 and a max value of 9 you could not pick a value that satisfies those constraints. I think that if you want to control the UI that closely you should use another mechanism to lay out the visualization for your customized application visualization.

wjwwood: I agree that having invertible sliders (e.g. 2-1-0 rather than 0-1-2) is desirable.

Then maybe we should consider different terminology, because minimum is less than or equal to maximum by definition in most situations. Maybe `from` and `to`?

tfoote: Ok, if we want an invertable slider, we should have a `bool invert_slider` in the description not in the constraint. However I don't believe that we should tangle that up in the constraint section. It's not related to the constraints. Constraint checking will have to be universally implemented and extra complexity there is not where we want things.

tfoote:
In #71 we talked about potentially supporting minimum and maximum constraints separately which I don't see supported here. @jacobperron suggested using arrays to make one end optional. Or we could encode them as separate constraints.

The other thing that we aren't supporting here is open intervals. I'm not sure that we need to but wanted to make sure it was a conscious decision.

wjwwood:
> In #71 we talked about potentially supporting minimum and maximum constraints separately which I don't see supported here.

And over there I suggested you make a pr to this pr to demonstrate what you mean. I didn't do it because I don't understand how it would work or what its purpose is.

> The other thing that we aren't supporting here is open intervals. I'm not sure that we need to but wanted to make sure it was a conscious decision.

What is an "open interval"?

rcl_interfaces/msg/FloatingPointRange.msg
wjwwood: I can buy the argument that we don't want extra complexity in the constraint enforcement, but on the other hand, you must check for this case anyways (we cannot prevent someone from passing a parameter descriptor which has a min > max), so you can choose to error in that case, or simply flip them for the purposes of validation.

I don't think we should embed something like `invert_slider` into the parameter descriptor. I see this discussion of using `from`/`to` instead as a way of changing what we're saying the constraint is in order to be more flexible, e.g. "the value is between the values from and to" versus "the value is greater than minimum and less than maximum".

This wouldn't work, however, if you wanted to bound just one side of the value, e.g. "greater than 5".  I suppose you could do what is done now, which is to select an arbitrarily large value. The other problem with bounding only one side is that rendering that as a slider is hard, because you don't know what scale to put the slider on.

My thinking is that these are actually two different kinds of constraints (mutually exclusive constraints but still different) which would be represented differently in a GUI. A range would be between two values, have a step size, and be represented by a slider. A "numeric constraint" would be minimum value or a maximum value and in the GUI that would simply be a text box that checked that the value entered didn't violate the constraint, i.e. no slider.

rcl_interfaces/msg/IntegerRange.msg
jubeira: If we use `from / to` and allow invertible sliders, should we expect the step be positive or negative when `from > to`?

rcl_interfaces/msg/FloatingPointRange.msg
jubeira: MHO:

> I can buy the argument that we don't want extra complexity in the constraint enforcement, but on the other hand, you must check for this case anyways (we cannot prevent someone from passing a parameter descriptor which has a min > max), so you can choose to error in that case, or simply flip them for the purposes of validation.

 :+1: to using `from` / `to` naming and to this rationale. The only warning I see is that this approach seems a bit more error prone than using `min` / `max` and not allowing `max` < `min`, but if using invertible sliders is desirable then it's worth it to gain flexibility.

> This wouldn't work, however, if you wanted to bound just one side of the value, e.g. "greater than 5". I suppose you could do what is done now, which is to select an arbitrarily large value. The other problem with bounding only one side is that rendering that as a slider is hard, because you don't know what scale to put the slider on.

I think using arbitrarily large numbers should be enough for most use-cases. In the end you are still constrained by the max (or min) value the datatype can hold. For floats you should be able to use `inf` right?

jubeira (commented):
I've left a couple of comments / questions; just to add one more opinion in general terms I'm aligned with this draft :+1:

> In #71 we talked about potentially supporting minimum and maximum constraints separately which I don't see supported here.

Isn't this what @wjwwood is saying in the last paragraph [here](https://github.com/ros2/rcl_interfaces/pull/72/files#r277116340) when talking about numeric constraints? If that's the case, it could be added as another constraint in the `ParameterDescriptor`.

> What is an "open interval"?

I think @tfoote is referring to constraints such as [1.0, 0.1, 5.0) i.e. excluding one of the bounds (please correct me if that's not the case). Considering the step is always discrete one way or another, this can be achieved by tweaking the bounds so as not to include the number in the border.
In other words, if I want a value to be between 1 and 5 with a step of 0.1, I can specify the constraint as [1.0, 0.1, 4.9] without needing any extra functionality.

rcl_interfaces/msg/IntegerRange.msg
gonzodepedro: How about having min and max. But using the sign in the step to represent direction?

dirk-thomas: > How about having min and max. But using the sign in the step to represent direction?

That sound like a good idea to me. :+1:

rcl_interfaces/msg/FloatingPointRange.msg
gonzodepedro: > I think using arbitrarily large numbers should be enough for most use-cases. In the end you are still constrained by the max (or min) value the datatype can hold. For floats you should be able to use inf right?

I agree

rcl_interfaces/msg/IntegerRange.msg
wjwwood: > If we use from / to and allow invertible sliders, should we expect the step be positive or negative when from > to?

The idea was to have step be a magnitude and therefore step would always be positive (unsigned). The relationship between from and to doesn't matter in that case.

> How about having min and max. But using the sign in the step to represent direction?

Personally I find that confusing, but I guess that's fine. You still need to check for min > max and do something in that case.

The nice thing about an unsigned step and from/to is that all of these instances are implicitly correct (`{from, to, step}`):

- `{0, 0, 0}`
- `{-1, 1, 0}`
- `{1, -1, 0}`
- `{-1, 1, 1}`
- `{1, 1, 1}`

And so on...

If you allow step to be > the distance between from and to, then there are no extra checks needed to consider it valid or not.

In the case of floating point you do have to additionally assert step is positive, unless you implicitly take the `fabs` of it always, in which case all values are possible again. You do have to have special logic to handle the direction of the slider, basically determining min/max based on from/to right before you render the GUI.

With min/max you need to additionally check that min < max and do different logic based on the sign of step instead of the ordering of from/to.

rcl_interfaces/msg/FloatingPointRange.msg
wjwwood: In the case of "greater than 5", I still don't think you want a slider, unless you pick some arbitrarily large, but not too large, upper bound like 100. If you use `inf` or `INT64_MAX`, then you probably end up with a useless slider. That's why I think they're two different kinds of constraints.

rcl_interfaces/msg/IntegerRange.msg
gonzodepedro: I agree. I'd got with from / to and step >= 0, 0 meaning no step.

gonzodepedro:
I think that moving on with from/to is ok. 
We can have a step >=0, 0 meaning no step.

gonzodepedro:
I created this PR #74 with changes based on the discussion on this PR.

rcl_interfaces/msg/ParameterDescriptor.msg
wjwwood: I guess we'll leave this as-is.

wjwwood:
Ok, I targeted this at master and rebased to resolve some conflicts and took it out of draft mode.

I'll let others push this through final review and CI and stuff, if that's ok.

rcl_interfaces/msg/ParameterDescriptor.msg
```diff
@@ -1,8 +1,32 @@
...
+# Description of the parameter, visible from introspection tools.
+string description
+
+# Parameter constraints
+
+# Plain English description of additional constraints which cannot be expressed
+# with the available constraints, e.g. "only prime numbers".
+# By convention, this should only be used to clarify constraints which cannot
+# be completely expressed with the parameter constraints below.
+string additional_constraints
```
dirk-thomas, line 17: I would argue the text can already be part of the description and doesn't have to be separate. E.g. a UI could show the description in a tooltip. I find it rather intuitive if it needs to concatenate two descriptions to provide all information.

wjwwood: The idea is that is only about constraints, and should be short. Where as the description would be longer and have information like what units it is in or how it used in the node.

I feel that it has value as a separate field. Consider that you're just using float range, you may still put in the text description that the parameter should be between 0.0 and 1.0. I don't see this as any different, you might want to speak about some of the constraints in the description, but repeat them more specifically or concisely in this field.

Also, I imagine that the GUI may not show the description all the time, perhaps only on hover, whereas this string would be shown always, maybe just under things like sliders or combo boxes as a footnote.

It would be nice to hear what others think about this, I did ask for feedback on it originally (https://github.com/ros2/rcl_interfaces/pull/73) and no one said anything.

jacobperron: I think there is a semantic difference between the describing a parameter and describing the constraints on a parameter. So having two fields makes sense to me.

On another note, I could see this field being abused (for better or worse) to actually do constraint checking in user-defined callbacks. For example, [in the Navigation WG meeting today](https://discourse.ros.org/t/ros2-navigation-wg-thursday-3-00-pm-pacific-gmt-7-00/7586/16), there was a discussion about having a mechanism for constraining parameter updates in certain states of a lifecycle node. This string could be used to encode a valid state. I'm not sure if this is something we should be worried about though.

wjwwood: I suppose they could abuse it that way, but they could also just put that into the description :x

That's only helpful, however, if the GUI understands it. So if you have cooperation with the GUI and code, why not just propose a new constraint for this message or just publish another message with your custom constraints...

---

W.r.t. the constraints being based on the lifecycle state, we should make it so that the ParameterDescriptor can be updated (currently you have to undeclare and re-declare a parameter to do that), so that they can simply update the constraints when the node itself changes states.

dirk-thomas: ROS 1 has range limits as well as a description for dynamic parameters. That has worked well in the past without a second text field describing the constraint in text form. Without a strong reason I don't think we should add a second description just based on the thoughts about how it could useful.

First get a minimal version merged and implemented from APIs in C++ and Python, do constraint checking, and have an rqt plugin to interact with these parameters and then consider extending it if there is a strong need / use case to do so.

wjwwood: I think the idea of how it should be used is sufficiently concrete to include it now, so I think we should keep it.

gonzodepedro:
I pushed a change that makes the double value in FloatingPointRange a float64 value, same type as used in ParameterValue.msg
I also added the new .msg files to the CMakeList

gonzodepedro:
CI
* Linux [Build Status](https://ci.ros2.org/job/ci_linux/6798/)
* Linux-aarch64 [Build Status](https://ci.ros2.org/job/ci_linux-aarch64/3095/)
* macOS [Build Status](https://ci.ros2.org/job/ci_osx/5590/)
* Windows [Build Status](https://ci.ros2.org/job/ci_windows/6624/)

rcl_interfaces/CMakeLists.txt
```diff
@@ -27,6 +27,8 @@ rosidl_generate_interfaces(${PROJECT_NAME}
   "msg/ParameterType.msg"
   "msg/ParameterValue.msg"
   "msg/SetParametersResult.msg"
+  "msg/FloatingPointRange.msg"
```
jacobperron: nitpick: alphabetize

jacobperron (approved):
LGTM! Would be good to get second approval though.

rcl_interfaces/msg/ParameterDescriptor.msg
wjwwood: To be clear here, I'm fine with removing it, if the consensus is to do that, but my personal opinion is that it should stay. Right now we've got a -1 from @dirk-thomas, +1 from me, and I'll call it a +0 from @jacobperron?

If anyone has a +/-1 opinion could comment here so we can have some more decisive conclusion would be nice.

wjwwood (commented):
I'd like to see a resolution on the conversation about `additional_constraints`, but otherwise lgtm. (cannot approve my own pr)

rcl_interfaces/msg/ParameterDescriptor.msg
jacobperron: +1

tfoote (approved):

PR #74, opened 2019-04-23
Changes based on comments on PR #72

gonzodepedro:


rcl_interfaces/msg/FloatingPointRange.msg
```diff
@@ -1,14 +1,13 @@
 # Represents bounds and a step value for a floating point typed parameter.
 
 # Lower bound for valid values, inclusive.
```
wjwwood: We should probably change this to `Left bound` and `Right bound` below instead of `Upper bound`.

rcl_interfaces/msg/FloatingPointRange.msg
```diff
@@ -1,14 +1,13 @@
...
-double maximum_value
+double to_value
 
-# Size of valid steps between the lower and upper bound.
+# Size of valid steps between the from and to bound.
 # Step should be positive, and a step of zero implies continuous values.
-# Step size should be less than the distance between the minimum and maximum
-# values, and ideally would be an even multiple of that distance, but that
-# is not required.
+# Step size should be less than the distance between the values, and ideally
```
wjwwood: I think we should relax this as @dirk-thomas suggested in order to allow negative step values (my proposal is just to take the `abs`/`fabs` of step always, so something like this (I extended the documentation further as well to cover more corner cases):

> Step is considered to be a magnitude, therefore negative values are treated the same as positive values, and a step value of zero implies a continuous range of values.
> Ideally, the step would be both less than or equal to the distance between the bounds, as well as an even multiple of the distance between the bounds, but neither are required.
> If the absolute value of the step is larger than or equal to the distance between the two bounds, then the bounds will be the only valid values, e.g. if the range is defined as `{from_value: 1.0, to_value: 2.0, step: 5.0}` then the valid values will be `1.0` and `2.0`.
> If the step is less than the distance between the bounds, but the distance is not a multiple of the step, then the "to" bound will always be a valid value, e.g. if the range is defined as `{from_value: 2.0, to_value: 5.0, step: 2.0}` then the valid values will be `2.0`, `4.0`, and `5.0`.

I think the nice thing about describing as above is that all combinations of values that you can put in the message are valid, even if they change behavior GUI or something like that in odd cases.

rcl_interfaces/msg/IntegerRange.msg
```diff
@@ -1,15 +1,13 @@
 # Represents bounds and a step value for an integer typed parameter.
 
 # Lower bound for valid values, inclusive.
```
wjwwood: Similar comment here, I think left and right are appropriate because we are speaking about a "number line" (https://en.wikipedia.org/wiki/Number_line).

rcl_interfaces/msg/IntegerRange.msg
```diff
@@ -1,15 +1,13 @@
...
-# Step should be positive, and a step of zero implies a step size of one, i.e.
-# all integer values between the bounds are valid.
-# Step size should be less than the distance between the minimum and maximum
-# values, and ideally would be an even multiple of that distance, but that
-# is not required.
-int64 step
+# Size of valid steps between the from and to bound.
+# Step should be positive, and a step of zero implies continuous values.
+# Step size should be less than the distance between the values, and ideally
+# would be an even multiple of that distance, but that is not required.
```
wjwwood: I think the suggested docs above apply here as well, but obviously need to be changed in a few ways for integers.

rcl_interfaces/msg/ParameterDescriptor.msg
```diff
@@ -19,10 +19,12 @@ string additional_constraints
...
 bool read_only false
 
-# docs...
+# If this bounded sequence contains a floating_point_range constraint it must
+# be enforced
 FloatingPointRange[<=1] floating_point_range
 
-# docs...
+# If this bounded sequence contains a integer_range constraint it must
+# be enforced
```
wjwwood: I would add that the range consists of a `from_value`, `to_value`, and a `step`, and that this and the float range are mutually exclusive.

rcl_interfaces/msg/FloatingPointRange.msg
gonzodepedro: I agree that Lower/Upper is not the right choice. Using left/right and lower/upper makes me think in minimums and maximums. 
My proposal would be to keep with the from/to idea and say: 

'Start value for valid values range' and 'End value for valid values range'

wjwwood: Sounds good to me.

rcl_interfaces/msg/FloatingPointRange.msg
jacobperron: +1 for explaining cases with examples. Though, I think based on our offline discussion yesterday, we don't want to include `to_value` if the distance is not a multiple of `step`.

wjwwood: Yes, that definitely needs to be changed.

rcl_interfaces/msg/ParameterDescriptor.msg
```diff
@@ -20,11 +20,13 @@ string additional_constraints
 bool read_only false
 
 # If this bounded sequence contains a floating_point_range constraint it must
-# be enforced
+# be enforced. FloatingPointRange consist of a from_value a to_value and a step.
```
wjwwood: nitpick: one sentence per line.

rcl_interfaces/msg/FloatingPointRange.msg
gonzodepedro: I don't fully understand why we don't want to linclude `to_value` if the distance is not a multiple of `step`.
The idea would be to remove it or to name it different?

jacobperron: The idea is that the value of `to_value` is not valid.
E.g. let `from_value=0`, `to_value=5`, and `step=2`
Then, valid values for the parameter are `[0, 2, 4]`.
`5` is not a valid value, even though it is the value of `to_value`.

jacobperron: The other case to document is `{from_value: 1.0, to_value: 2.0, step: 5.0}`

I'm not sure if we discussed this, but I think applying the same rule we are left with a single valid parameter value: `1.0`.

jubeira: > I don't fully understand why we don't want to linclude to_value if the distance is not a multiple of step.

From yesterday's discussion, I understand that it makes more sense and it's more clear to constraint the values to {~to_value~ `from_value` +/- n * `step`). In general terms you should be able to adjust `to_value` to comply with that constraint.

At least to me, in cases like:
- `from_value` = 0
- `step` = 4
- `to_value` = 5

I really don't see a practical use of including `to_value`. And if the step size is larger than `|to_value - from_value|`, then only ~to_value~ `from_value` should be allowed.

jacobperron: @jubeira I think that is almost correct. Since values go "from" -> "to", I think you mean to use {`from_value` +- n * `step`} as the constraint.

And,

```diff
- And if the step size is larger than `|to_value - from_value|`, then only `to_value` should be allowed.
+ And if the step size is larger than `|to_value - from_value|`, then only `from_value` should be allowed.
```

jubeira: Thanks @jacobperron, yes, that's exactly what I meant!

gonzodepedro: What I have in my mind is that we state in the documentation that there is a `from_value` a `to_value` and a `step`. Then we state `Step size should be less than the distance between the values, and a multiple of that distance`. If those conditions are not met I think that the constraint as a whole should be rejected because it doesn't satisfy the required conditions.

rcl_interfaces/msg/ParameterDescriptor.msg
```diff
@@ -19,10 +19,16 @@ string additional_constraints
 # If 'true' then the value cannot change after it has been initialized.
 bool read_only false
 
-# docs...
+# If this bounded sequence contains a floating_point_range constraint it must
+# be enforced.
+# FloatingPointRange consist of a from_value a to_value and a step.
```
wjwwood: `consists`?

wjwwood: Also, use comma's:

> A FloatingPointRange consists of a `from_value`, a `to_value`, and a `step`.

rcl_interfaces/msg/ParameterDescriptor.msg
```diff
@@ -19,10 +19,16 @@ string additional_constraints
...
+# FloatingPointRange and IntegerRange are mutually exclusive
 FloatingPointRange[<=1] floating_point_range
 
-# docs...
+# If this bounded sequence contains a integer_range constraint it must
+# be enforced.
+# IntegerRange consist of a from_value a to_value and a step.
+# FloatingPointRange and IntegerRange are mutually exclusive
```
wjwwood: Punctuation.

rcl_interfaces/msg/FloatingPointRange.msg
wjwwood: > If those conditions are not met I think that the constraint as a whole should be rejected because it doesn't satisfy the required conditions.

We didn't agree to that, as far as I understood our last meeting.

wjwwood: Also you still need `to_value` in the case that `step` is `0`.

wjwwood: > The other case to document is {from_value: 1.0, to_value: 2.0, step: 5.0}

> I'm not sure if we discussed this, but I think applying the same rule we are left with a single valid parameter value: 1.0.

I think we didn't decide on this either, but I'm not sure I agree with that conclusion. This is a consequence of deciding that `to_value` is not always a valid value, which was not what I was thinking in my original proposal, but is what we decided during the meeting.

rcl_interfaces/msg/FloatingPointRange.msg
```diff
@@ -1,14 +1,13 @@
 # Represents bounds and a step value for a floating point typed parameter.
 
-# Lower bound for valid values, inclusive.
-double minimum_value
+# Start value for valid values range, inclusive.
```
wjwwood: This looks weird to me, maybe instead:

> Start value for valid values in the range, inclusive.

Or maybe even better:

> Start value for the range of valid values, inclusive.

Honestly the original:

> Start value for valid values, inclusive.

Was probably the best.

rcl_interfaces/msg/FloatingPointRange.msg
```diff
@@ -1,14 +1,13 @@
...
+double to_value
 
-# Size of valid steps between the lower and upper bound.
+# Size of valid steps between the from and to bound.
 # Step should be positive, and a step of zero implies continuous values.
-# Step size should be less than the distance between the minimum and maximum
-# values, and ideally would be an even multiple of that distance, but that
-# is not required.
+# Step size should be less than the distance between the values.
+# The to_value is valid even if the step is not a multiple of it.
```
wjwwood: What happens if step is larger than the distance?

What about the examples from my proposed doc block (https://github.com/ros2/rcl_interfaces/pull/74#discussion_r277856900)?

rcl_interfaces/msg/IntegerRange.msg
```diff
@@ -1,15 +1,13 @@
...
 
-# Size of valid steps between the lower and upper bound.
-# Step should be positive, and a step of zero implies a step size of one, i.e.
-# all integer values between the bounds are valid.
-# Step size should be less than the distance between the minimum and maximum
-# values, and ideally would be an even multiple of that distance, but that
-# is not required.
-int64 step
+# Size of valid steps between the from and to bound.
+# Step should be positive, and a step of zero implies continuous values.
```
wjwwood: > Step should be positive

Redundant, it is unsigned.

rcl_interfaces/msg/FloatingPointRange.msg
```diff
@@ -1,14 +1,13 @@
...
+# End value for valid values range, inclusive.
+double to_value
 
-# Size of valid steps between the lower and upper bound.
+# Size of valid steps between the from and to bound.
 # Step should be positive, and a step of zero implies continuous values.
-# Step size should be less than the distance between the minimum and maximum
-# values, and ideally would be an even multiple of that distance, but that
-# is not required.
+# Step size should be less than the distance between the values.
```
wjwwood: less than or equal to?

rcl_interfaces/msg/ParameterDescriptor.msg
```diff
@@ -19,10 +19,16 @@ string additional_constraints
 # If 'true' then the value cannot change after it has been initialized.
 bool read_only false
 
-# docs...
+# If this bounded sequence contains a floating_point_range constraint it must
+# be enforced.
```
wjwwood: I don't think saying it must be enforced is correct. The user may choose to enforce it strictly, conditionally, or not at all for various scenarios.

wjwwood: I'd word it more like:

> If this sequence is not empty, the constraint inside of it applies to this parameter.

And honestly, that could be said once in this file as it applies the same to all optional constraints.

gonzodepedro:
I think I addressed all the comments based on what was discussed here and offline.

rcl_interfaces/msg/IntegerRange.msg
```diff
@@ -1,13 +1,22 @@
...
+# bounds, as well as an even multiple of the distance between the bounds, but
+# neither are required.
+# If the absolute value of the step is larger than or equal to the distance
+# between the two bounds, then the bounds will be the only valid values,
+# e.g. if the range is defined as {from_value: 1.0, to_value: 2.0, step: 5.0} then
+# the valid values will be 1.0 and 2.0.
+# If the step is less than the distance between the bounds, but the distance is
+# not a multiple of the step, then the "to" bound will always be a valid value,
+# e.g. if the range is defined as {from_value: 2.0, to_value: 5.0, step: 2.0} then
+# the valid values will be 2.0, 4.0, and 5.0.
```
wjwwood: Please use integers for the examples, as this is the integer range.

rcl_interfaces/msg/IntegerRange.msg
```diff
@@ -1,13 +1,22 @@
...
-# End value for valid values range, inclusive.
+# End value for valid values, inclusive.
 int64 to_value
 
 # Size of valid steps between the from and to bound.
-# Step should be positive, and a step of zero implies continuous values.
-# Step size should be less than the distance between the values.
-# The to_value is valid even if the step is not a multiple of it.
+# A step value of zero implies a continuous range of values.
+# Ideally, the step would be both less than or equal to the distance between the
```
wjwwood: I know this is based on my suggestion originally, but now that I read it again, I think we should drop the word `both`, because `as well as` is sufficient and `both less than or equal` is a bit confusing on the first read.

rcl_interfaces/msg/FloatingPointRange.msg
```diff
@@ -1,9 +1,9 @@
...
 
-# Start value for valid values range, inclusive.
+# Start value for valid values, inclusive.
 double from_value
 
-# End value for valid values range, inclusive.
+# End value for valid values, inclusive.
 double to_value
 
 # Size of valid steps between the from and to bound.
```
wjwwood, line 9: The float range's documentation of step was not updated.

wjwwood:
> I think I addressed all the comments based on what was discussed here and offline.

I think there are still several that have not been addressed.

rcl_interfaces/msg/ParameterDescriptor.msg
```diff
@@ -19,16 +19,16 @@ string additional_constraints
...
 
-# If this bounded sequence contains a floating_point_range constraint it must
-# be enforced.
-# FloatingPointRange consist of a from_value a to_value and a step.
-# FloatingPointRange and IntegerRange are mutually exclusive
+
+# If any of the following sequences are not empty, then the constraint inside of
+# them apply to this parameter.
+
+# FloatingPointRange consists of a from_value, a to_value and a step.
```
wjwwood: Please include the Oxford comma (it's a subjective choice, not right or wrong, but we do include it elsewhere in our code base):

>  FloatingPointRange consists of a from_value, a to_value, and a step.

rcl_interfaces/msg/ParameterDescriptor.msg
```diff
@@ -19,16 +19,16 @@ string additional_constraints
...
+# FloatingPointRange and IntegerRange are mutually exclusive.
 FloatingPointRange[<=1] floating_point_range
 
-# If this bounded sequence contains a integer_range constraint it must
-# be enforced.
-# IntegerRange consist of a from_value a to_value and a step.
-# FloatingPointRange and IntegerRange are mutually exclusive
+# IntegerRange consists of a from_value, a to_value and a step.
```
wjwwood: Here too.

wjwwood (approved):
After the minor style change is addressed, lgtm!

Thanks for pushing on it, @gonzodepedro.

jacobperron (approved):
LGTM!

PR #76, opened 2019-04-26
Add fields for response code to action_msgs/CancelGoal service

### LINKED ISSUE #63: Add response code to action_msgs/CancelGoal service
#### Feature description

It was discussed as part of the [Actions design doc](https://github.com/ros2/design/pull/193#discussion_r228721050) that a response code be added to the `CancelGoal` service to better communicate to an action client the reason for a declined cancel request.
Currently, in the event of a declined request (whether it is due to invalid data or is flat-out rejected by the server) the only information the action client receives is an empty list of goal IDs.

#### Implementation considerations

To avoid confusion with the action Result service, I propose we avoid the word "result" as proposed in https://github.com/ros2/design/pull/193#discussion_r229529502 and instead use "response":

```
int8 RESPONSE_OK = 0
int8 RESPONSE_REJECTED = 1
# A specific goal was attempted to be canceled, but no such goal exists
int8 RESPONSE_INVALID_GOAL_ID = 2

# Goal info containing an ID and timestamp
GoalInfo goal_info
---
# Goals that accepted the cancel request
GoalInfo[] goals_canceling
# Response code
int8 response
```

Open to ideas on changes to the possible error codes.

jacobperron:
Resolves #63 

Useful for communicating the reason for not accepting a cancel goal request.

### DIFF
--- a/action_msgs/srv/CancelGoal.srv
+++ b/action_msgs/srv/CancelGoal.srv
@@ -11,25 +11,25 @@
 # Goal info containing an ID and timestamp
 GoalInfo goal_info
 ---
-# Response codes
-# Indicates the request was accepted.
+# Return codes
+# Indicates the request was accepted without any errors.
 # One or more goals have transitioned to the CANCELING state.
 # The goals_canceling list is not empty.
-int8 RESPONSE_OK=0
+int8 ERROR_NONE=0
 # Indicates the request was rejected.
 # No goals have transitioned to the CANCELING state.
 # The goals_canceling list is empty.
-int8 RESPONSE_REJECTED=1
+int8 ERROR_REJECTED=1
 # Indicates the requested goal ID does not exist.
 # No goals have transitioned to the CANCELING state.
 # The goals_canceling list is empty.
-int8 RESPONSE_UNKNOWN_GOAL_ID=2
+int8 ERROR_UNKNOWN_GOAL_ID=2
 # Indicates the goal is not cancelable because it is already in a terminal state.
 # No goals have transitioned to the CANCELING state.
 # The goals_canceling list is empty.
-int8 RESPONSE_GOAL_TERMINATED=3
+int8 ERROR_GOAL_TERMINATED=3
 
-# Reponse code
-int8 response
+# Return code
+int8 return_code
 # Goals that accepted the cancel request
 GoalInfo[] goals_canceling
--- a/action_msgs/srv/CancelGoal.srv
+++ b/action_msgs/srv/CancelGoal.srv
@@ -20,12 +20,14 @@ int8 RESPONSE_OK=0
 # No goals have transitioned to the CANCELING state.
 # The goals_canceling list is empty.
 int8 RESPONSE_REJECTED=1
-# Indicates the requested goal ID is invalid.
-# This could be because a goal with the ID is not cancelable (e.g. terminated)
-# or there is no goal that matches the ID.
+# Indicates the requested goal ID does not exist.
 # No goals have transitioned to the CANCELING state.
 # The goals_canceling list is empty.
-int8 RESPONSE_INVALID_GOAL_ID=2
+int8 RESPONSE_UNKNOWN_GOAL_ID=2
+# Indicates the goal is not cancelable because it is already in a terminal state.
+# No goals have transitioned to the CANCELING state.
+# The goals_canceling list is empty.
+int8 RESPONSE_GOAL_TERMINATED=3
 
 # Reponse code
 int8 response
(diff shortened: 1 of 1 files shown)


action_msgs/srv/CancelGoal.srv
```diff
@@ -11,5 +11,23 @@
...
 GoalInfo goal_info
 ---
+# Response codes
+# Indicates the request was accepted.
+# One or more goals have transitioned to the CANCELING state.
+# The goals_canceling list should not be empty.
+int8 RESPONSE_OK=0
+# Indicates the request was rejected.
+# No goals have transitioned to the CANCELING state.
+# The goals_canceling list should be empty.
```
dirk-thomas: Consistency: here it says "should be empty" while for the next response code it says "is empty".

action_msgs/srv/CancelGoal.srv
```diff
@@ -11,5 +11,23 @@
...
+int8 RESPONSE_REJECTED=1
+# Indicates the requested goal ID is invalid.
+# This could be because a goal with the ID is not cancelable (e.g. terminated)
+# or there is no goal that matches the ID.
+# No goals have transitioned to the CANCELING state.
+# The goals_canceling list is empty.
+int8 RESPONSE_INVALID_GOAL_ID=2
+
+# Reponse code
+int8 response
```
dirk-thomas: A `response` field in the response message sounds awkward. Maybe rename the flag to `status` (or similar - not sure what other messages use).

action_msgs/srv/CancelGoal.srv
```diff
@@ -11,5 +11,23 @@
...
+# One or more goals have transitioned to the CANCELING state.
+# The goals_canceling list should not be empty.
+int8 RESPONSE_OK=0
+# Indicates the request was rejected.
+# No goals have transitioned to the CANCELING state.
+# The goals_canceling list should be empty.
+int8 RESPONSE_REJECTED=1
+# Indicates the requested goal ID is invalid.
+# This could be because a goal with the ID is not cancelable (e.g. terminated)
+# or there is no goal that matches the ID.
```
dirk-thomas: Are there other reasons (than terminated) for not being cancelable? If yes, maybe they would benefit from separate codes.

jacobperron: Not that I can think of. I can make them separate error codes: `UNKNOWN_GOAL_ID` and `GOAL_TERMINATED`

action_msgs/srv/CancelGoal.srv
jacobperron: b5d202d

action_msgs/srv/CancelGoal.srv
jacobperron: 29eb7ff

action_msgs/srv/CancelGoal.srv
jacobperron: I'll take a look around for some precedent, but I'm hesitant to use `status` so it's not confused for the goal status. The original proposal was to use `result`, which I didn't want to confuse with the goal result, but I think this is sufficiently removed from that context that it could work.

jacobperron: What about `error_code`?

```
ERROR_NONE=0
ERROR_REJECTED=1
ERROR_UNKNOWN_GOAL_ID=2
ERROR_GOAL_TERMINATED=3
```

action_msgs/srv/CancelGoal.srv
jacobperron: \+ 62eceb4

action_msgs/srv/CancelGoal.srv
wjwwood: Or `return_code`?

dirk-thomas: > Or `return_code`?

Sounds good to me - especially with the semantic `ERROR_NONE=0`.

jacobperron: Updated 1591668

dirk-thomas (approved):

jacobperron:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/6816/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/3111/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/5605/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/6639/)

CI is also for the following related PRs:
* https://github.com/ros2/rcl/pull/422
* https://github.com/ros2/rclcpp/pull/710

PR #82, opened 2019-06-05
use idl files (if any)

dirk-thomas:
Related to ros2/test_interface_files#6.


jacobperron (commented):
With the current state of https://github.com/ros2/test_interface_files/pull/6, this will not build the IDL messages. To help prevent regressions, is it possible to only generate the new message for fastrtps?

dirk-thomas:
> is it possible to only generate the new message for fastrtps?

Where and how should that condition be added?

jacobperron:
> Where and how should that condition be added?

It would be nice if we could do it in the CMakeLists.txt, but I'm not sure how. If there's not a known easy way to do this then we can leave it as-is.

jacobperron (approved):

PR #90, opened 2020-03-20
Remove unused ParameterEventDescriptors.msg

brawner:
This message file was added several years ago and is not currently in use. Instead of adding documentation for an unused feature, I figured it would be better to just remove the message for now.

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/9708/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/5364/)
* macOS [Build Status](http://ci.ros2.org/job/ci_osx/7913/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/9602/)

### DIFF
--- a/rcl_interfaces/CMakeLists.txt
+++ b/rcl_interfaces/CMakeLists.txt
@@ -23,7 +23,6 @@ rosidl_generate_interfaces(${PROJECT_NAME}
   "msg/ListParametersResult.msg"
   "msg/Log.msg"
   "msg/ParameterDescriptor.msg"
-  "msg/ParameterEventDescriptors.msg"
   "msg/ParameterEvent.msg"
   "msg/Parameter.msg"
   "msg/ParameterType.msg"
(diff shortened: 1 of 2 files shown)


brawner:
@tfoote This was message I was talking about. We can also leave it in if that's preferable.

tfoote (changes requested):
This message was part of the design to provide a way for UIs to subscribe to a lighter weight version of the ParameterEvents topic. In particular they would be expected to subscribe to this and then only call getParameter for the ones that they care about. 

For most parameters this isn't a big deal, but for large parameters with lots of data this will significantly reduce the volume. As well as for tools that may want to support reconfiguring anything, but might only have a few parameters actively in the configuration gui this lets them be responsive, but not hog bandwidth. We ran out of time to implement that and didn't have the demand for it yet. but I think we should instead plan to implement that rather than removing this message.

brawner:
Ok, good points. I'll add information about it to the documentation PR, and maybe a one-liner that the feature has not yet been implemented. I have no issue with it sticking around if you still think it will be part of future functionality.

PR #101, opened 2020-04-23
Add new msgs for use in rclcpp/test

prajakta-gokhale:
Add a dummy message with `Header` and `Time` fields for use in rclcpp unit testing.

Related to https://github.com/ros2/rclcpp/pull/1080

Signed-off-by: Prajakta Gokhale <prajaktg@amazon.com>

### DIFF
--- a/test_msgs/CMakeLists.txt
+++ b/test_msgs/CMakeLists.txt
@@ -26,6 +26,8 @@ rosidl_generate_interfaces(${PROJECT_NAME}
   ${test_interface_files_ACTION_FILES}
   ${test_interface_files_IDL_FILES}
   "msg/Builtins.msg"
+  "msg/Header.msg"
+  "msg/MessageWithHeader.msg"
   "action/NestedMessage.action"
   DEPENDENCIES builtin_interfaces
   ADD_LINTER_TESTS
--- a/test_msgs/msg/Header.msg
+++ b/test_msgs/msg/Header.msg
@@ -1,0 +1,3 @@
+# Simple Header message with a timestamp field.
+
+builtin_interfaces/Time stamp
--- a/test_msgs/msg/MessageWithHeader.msg
+++ b/test_msgs/msg/MessageWithHeader.msg
@@ -1,0 +1,3 @@
+# Message containing a simple Header field.
+
+Header header


wjwwood:
This seems reasonable, @jacobperron what do you think?

Is this enough to trick your code into considering it? I assumed (for some reason) it had to literally be the type `std_msgs/msg/Header`.

prajakta-gokhale:
> Is this enough to trick your code into considering it? I assumed (for some reason) it had to literally be the type `std_msgs/msg/Header`.

This seems to be working locally for me at least! I guess [this code](https://github.com/ros-tooling/libstatistics_collector/blob/master/include/libstatistics_collector/topic_statistics_collector/received_message_age.hpp#L43-L48) works as long as there is _a_ header present in the message.

jacobperron:
I was hoping to figure out what we're missing from the CMakeLists.txt in rclcpp (https://github.com/ros2/rclcpp/pull/1080#issuecomment-618174931), since I think it's a better solution.

This solution is okay, but we need to account for the newly added messages in several places. For example, `test_communication` will likely fail in several places because it doesn't expect these messages to exist ([for example](https://github.com/ros2/system_tests/blob/master/test_communication/test/test_publisher.cpp#L99)).  Adding more interfaces to `test_msgs` may significantly increase our integration test time (I'm not sure), or we can update the integration tests to ignore unknown test messages.

prajakta-gokhale:
> I was hoping to figure out what we're missing from the CMakeLists.txt in rclcpp ([ros2/rclcpp#1080 (comment)](https://github.com/ros2/rclcpp/pull/1080#issuecomment-618174931)), since I think it's a better solution.

I agree, but I wasn't able to figure out what was wrong earlier. If you have any pointers I'd be happy to try again.

> This solution is okay, but we need to account for the newly added messages in several places. For example, `test_communication` will likely fail in several places because it doesn't expect these messages to exist ([for example](https://github.com/ros2/system_tests/blob/master/test_communication/test/test_publisher.cpp#L99)).

I started CI for all [`--packages-above rclcpp`](https://github.com/ros2/rclcpp/pull/1080#issuecomment-618197163) to make sure nothing else breaks.

>  Adding more interfaces to `test_msgs` may significantly increase our integration test time (I'm not sure), or we can update the integration tests to ignore unknown test messages.

I can do that as a follow-up PR.

prajakta-gokhale:
We found a way to do this w/o this change in https://github.com/ros2/rclcpp/pull/1080#issuecomment-618678820. No longer needed.

PR #105, opened 2020-05-19
Add action_msgs::msg::GoalStatus::STATUS_PREEMPTED

naiveHobo:
Signed-off-by: Sarthak Mittal <sarthakmittal2608@gmail.com>

As per https://github.com/ros2/rclcpp/issues/1104 and the corresponding PR https://github.com/ros2/rclcpp/pull/1117, this PR adds `STATUS_PREEMPTED` to `action_msgs/GoalStatus` to refer to a goal that was preempted by the client.

### DIFF
--- a/action_msgs/msg/GoalStatus.msg
+++ b/action_msgs/msg/GoalStatus.msg
@@ -25,6 +25,9 @@ int8 STATUS_CANCELED  = 5
 # The goal was terminated by the action server without an external request.
 int8 STATUS_ABORTED   = 6
 
+# The goal was preempted by the client.
+int8 STATUS_PREEMPTED = 7
+
 # Goal info (contains ID and timestamp).
 GoalInfo goal_info
 


SteveMacenski (approved):

jacobperron:
I don't think it's as simple as adding a new field to the message, there is also code in `rcl_action` and `rclpy` that works with GoalStatus. Before making further changes, I suggest a discussion and changes around the design of the state machine (see https://github.com/ros2/rclcpp/issues/1104#issuecomment-631737468)

fujitatomoya (approved):
This fix itself is LGTM.

SteveMacenski:
@jacobperron mentions that it may be worth adding the PREEMPTING as well - I agree as long as the intention is to add that to the state machine.

SteveMacenski:
@naiveHobo I think the best route here is to add the `PREEMPTING` message he asked for so that these messages are complete and then work with Jacob to update the state machine. Given we had one like this for ROS1, it shouldn't be too big of an undertaking. Once we have the design and this message in place, we can get the `rclcpp` PR merged.

PR #109, opened 2020-07-10
test_msgs: remove dependency action_msgs from package.xml?

meyerj:
This pull request is more a question than an actual request to merge my patch: Why is the dependency on `action_msgs` needed? `action_msgs` is not mentioned anywhere within `test_msgs`. Or must any interface package that defines actions declare a dependency on `action_msgs`?

This seems to be related to https://github.com/ros2/rosidl/pull/369, https://github.com/ros2/rcl_interfaces/pull/81 and other pull requests following https://github.com/ros2/rosidl/pull/369. In my understanding the explicit `action_msgs` dependency in `package.xml` should have been removed, too. But then the build can fail if the respective package is in the same workspace as `action_msgs`? I assume it just does not matter for most packages which are not part of the core ROS 2 distribution and are only ever built in an overlay workspace, and even if not it is unlikely to trigger a build error because `action_msgs` is built before most other packages anyway due to lexical ordering.

If `<depend>action_msgs</depend>` is indeed required for all packages which have actions, `rosidl_generate_interfaces()` should probably warn if that is not the case even if `find_package(action_msgs)` succeeds? I also could not find a hint about `action_msgs` in user-facing documentation or tutorials, e.g. [About ROS 2 interfaces](https://index.ros.org/doc/ros2/Concepts/About-ROS-Interfaces/), [Expanding on ROS 2 interfaces](https://index.ros.org/doc/ros2/Tutorials/Single-Package-Define-And-Use-Interface/) or [Understanding ROS 2 actions](https://index.ros.org/doc/ros2/Tutorials/Understanding-ROS2-Actions/).

I noticed when working on https://github.com/orocos/rtt_ros2_integration/pull/12, where the dependencies of an interface package are translated to dependencies of the respective generated Orocos typekit package. `test_msgs` is used for unit tests there.

### DIFF
--- a/test_msgs/package.xml
+++ b/test_msgs/package.xml
@@ -14,7 +14,6 @@
   <build_depend>builtin_interfaces</build_depend>
   <build_depend>test_interface_files</build_depend>
 
-  <depend>action_msgs</depend>
   <exec_depend>builtin_interfaces</exec_depend>
   <exec_depend>rosidl_default_runtime</exec_depend>
 


jacobperron:
> Or must any interface package that defines actions declare a dependency on action_msgs?

Code generated from an action definition (a `.action` file) makes use of the definitions in `action_msgs`, so it's best to declare a dependency to `action_msgs`.
This is briefly mentioned in the [Creating an Action tutorial](https://index.ros.org/doc/ros2/Tutorials/Actions/Creating-an-Action/).

I would prefer if we didn't have to depend on `action_msgs` at all, but we've currently got a "chicken-egg" problem: we depend on `action_msgs` as part of generating action interfaces (`rosidl_cmake`), but `action_msgs` itself depends on `rosidl_cmake` for the code generation pipeline. The logic added in https://github.com/ros2/rosidl/pull/369 was a compromise, requiring the user to declare the dependency on `action_msgs`, with a CMake error if they don't. It's true if `action_msgs` is already installed (or built) as a transitive dependency, then you don't *need* to declare it as a dependency, but it's probably best to do it just in case.

> If `<depend>action_msgs</depend>` is indeed required for all packages which have actions, `rosidl_generate_interfaces()` should probably warn if that is not the case even if `find_package(action_msgs)` succeeds?

A warning would be nice. I think this case is similar to forgetting to declare any dependency in the package.xml and having `find_package` work because the dependency was brought in transitively. I guess the only difference with `action_msgs` is that we don't require the user to have to `find_package` it. I would be open to reviewing a PR that adds such a warning for `action_msgs`.

I'd be even more excited about ideas for a solution involving the removal of the circular dependency between `action_msgs` and `rosidl_cmake`. I haven't spent time thinking about it in more detail, but I imagine there must be a way.

jacobperron:
Thinking about it some more, I think there's one pro for requiring users to declare an explicit dependency on `action_msgs` if they want code generation support for actions. This way users don't pay for what they don't use. I.e. if someone doesn't want to use actions then they don't need to build/install `action_msgs` on their system.

I'm going to close this, but feel free to continue the discussion here.

PR #115, opened 2020-11-05
Add boolean in ParameterDescriptor to enforce parameters static typing

ivanpauno:
Currently parameters can change of type any time, which can be avoided by using a custom parameter callback (e.g. [here](https://github.com/ros2/rclcpp/blob/c88cc649d32cb3a0ca7eb3c2b92d678303adb0d5/rclcpp/src/rclcpp/time_source.cpp#L104)).
When a parameter is changed to an unexpected type, it will usually generate an exception in the code using the parameter (see for example https://github.com/ros2/rclcpp/issues/802), which is undesired.

What users commonly expect is that a parameter not change its type, so it would be good to provide an easy way to ensure that.
The parameter descriptor already contains a [type](https://github.com/ros2/rcl_interfaces/blob/7a6ebaa2064151a48dd189cd514fc7db5d260047/rcl_interfaces/msg/ParameterDescriptor.msg#L3-L7) field, but that is being used to reflect the last type that was set.

The proposal of this PR is to add a field called `static_typing`, when `true` a parameter cannot be changed of type.
In that case, after the parameter is set to a type different than `PARAMETER_NOT_SET`, that type is not allowed to change.
Open questions:
- Should it be possible to declare a parameter with `static_typing=true` and `type=PARAMETER_NOT_SET`?
  I would say no, an `int` parameter should have an int default value provided when declared, and the user override (`--ros-args -p ...`) should match the type of the default value. i.e.:
  ```cpp
  // should raise, initialized with type PARAMETER_NOT_SET and requested static typing
  node->declare_parameter("an_int", rclcpp::ParameterValue(), descriptor_static_typing_true);
  // Okey, and it raises an exception if the user provides an override which is not an integer, e.g.: `-p an_int:=foo`
  node->declare_parameter("an_int", 5, descriptor_static_typing_true);
- Should it be possible to undeclare a parameter when `static_typing=true`? IMO, no.
  Allowing that would actually make a parameter of type `int` behave as an `optional<int>`, and it would still require special care when getting the parameter to avoid errors.

Alternatives:

- Add a `uint8 allowed_type` field in the parameter description, where `PARAMETER_NOT_SET` would allow dynamic typing. There's no big difference with the proposal above.
- Use the old `type` field as the `allowed_type` field described above.
- Add a `uint8[<=10] allowed_types` field (we could use a boolean mask instead of an array). This approach is more flexible, but 99% of the time the user just wants to allow a single type, so not supporting this case doesn't sound important (and can still be achieved with dynamic typing and an user provided parameter callback).

### DIFF
--- a/rcl_interfaces/msg/ParameterDescriptor.msg
+++ b/rcl_interfaces/msg/ParameterDescriptor.msg
@@ -21,8 +21,8 @@ string additional_constraints
 # If 'true' then the value cannot change after it has been initialized.
 bool read_only false
 
-# If 'true', duck typing is allowed.
-bool allow_duck_typing false
+# If 'true', the parameter is not allowed to change its type.
+bool static_typing true
 
 # If any of the following sequences are not empty, then the constraint inside of
 # them apply to this parameter.
--- a/rcl_interfaces/msg/ParameterDescriptor.msg
+++ b/rcl_interfaces/msg/ParameterDescriptor.msg
@@ -21,6 +21,8 @@ string additional_constraints
 # If 'true' then the value cannot change after it has been initialized.
 bool read_only false
 
+# If 'true', duck typing is allowed.
+bool allow_duck_typing false
 
 # If any of the following sequences are not empty, then the constraint inside of
 # them apply to this parameter.


ivanpauno:
https://github.com/ros2/rclcpp/pull/1436 implements the above proposal except this point:

> Should it be possible to declare a parameter with static_typing=true and type=PARAMETER_NOT_SET?
> I would say no,

Though adding that is easy.

rcl_interfaces/msg/ParameterDescriptor.msg
```diff
@@ -21,6 +21,8 @@ string additional_constraints
 # If 'true' then the value cannot change after it has been initialized.
 bool read_only false
 
+# If 'true', the parameter is not allowed to change its type.
+bool static_typing true
```
clalancette, line 25: Maybe:

```suggestion
bool statically_typed true
```

clalancette (commented):
Along with the other code to implement this, I think this is fine.

My one hesitation here is that making this change means that Galactic will not be able to communicate with Foxy and earlier (at least, for `ParameterDescriptor`).  We don't typically guarantee that, but it does happen to mostly work between Dashing, Eloquent, and Foxy.  That shouldn't prevent this from going in, but it is something we should note in the release notes.

ivanpauno:
> My one hesitation here is that making this change means that Galactic will not be able to communicate with Foxy and earlier (at least, for ParameterDescriptor)

Yeah, wire compatibility is hard in this case.
I think that the only way to achieve that here would be to reuse the `type` field which is already part of the parameter description.

gbiggs:
> Should it be possible to declare a parameter with static_typing=true and type=PARAMETER_NOT_SET?
I would say no

I agree.

> Should it be possible to undeclare a parameter when static_typing=true? IMO, no.

I agree with this, too.

clalancette:
> Should it be possible to declare a parameter with static_typing=true and type=PARAMETER_NOT_SET?
> I would say no

While I agree with this in principle, I think this will cause a lot of downstream headaches.  For instance, I know that navigation2 does things like this: https://github.com/ros-planning/navigation2/blob/11b25a51981eb53db6a6a3601729ae85238d9241/nav2_dwb_controller/dwb_core/src/dwb_local_planner.cpp#L79 .  That declares a parameter as NOT_SET by default, but expects the user to explicitly set this.  I *think* that would run afoul of this limitation.  So I think this is one we'll have to allow for pragmatic reasons.

> Should it be possible to undeclare a parameter when static_typing=true? IMO, no.

I agree with this one.

clalancette (commented):
Overall, I agree with the idea and changing this descriptor.  The details to sort out in my mind:

1.  The name of the parameter (I like my suggested name slightly better)
1.  What to do about allowing NOT_SET -> <TYPE> conversion
1.  The actual implementation.
1.  Adding some documentation to the Galactic release notes about this semantic change.

ivanpauno:
> While I agree with this in principle, I think this will cause a lot of downstream headaches. For instance, I know that navigation2 does things like this: https://github.com/ros-planning/navigation2/blob/11b25a51981eb53db6a6a3601729ae85238d9241/nav2_dwb_controller/dwb_core/src/dwb_local_planner.cpp#L79 .

That doesn't seem like a problem, as they are already "failing" [when a default value is not provided](https://github.com/ros-planning/navigation2/blob/bf22f34cc805d29361934b57b1791e9fb10a47a6/nav2_util/include/nav2_util/node_utils.hpp#L103-L105).
So it seems to me that they are never declaring a parameter without a default value (i.e. type = NOT_SET).

ivanpauno:
> The name of the parameter (I like my suggested name slightly better)

That sounds good to me.

> Adding some documentation to the Galactic release notes about this semantic change.

Probably, it's worth creating a thread in ros discourse, though I would like first to have an implementation working with all the details sorted out.

ivanpauno:
@jacobperron friendly ping, what are your thoughts about this?

jacobperron:
> That doesn't seem like a problem, as they are already "failing" when a default value is not provided.
So it seems to me that they are never declaring a parameter without a default value (i.e. type = NOT_SET).

I'm a bit confused by this point. I think the concern is valid. The nav stack [defaults several parameters to NOT_SET](https://github.com/ros-planning/navigation2/blob/8d6cd2878bae389d932982e759acb760f3291f63/nav2_util/include/nav2_util/node_utils.hpp#L87) and then is checking that the use provides a value. So, for this type of use-case, users will not be able to use statically typed parameters (unless we allow having an initial NOT_SET state). If users were able to detect if a default value was overridden, then they could use that signal as a workaround.

jacobperron:
Perhaps it's okay to disallow statically typed parameters that default to NOT_SET, but I think we want to be sure in that choice, since it looks like changing our minds later will require another change to `ParameterDescriptor.msg`.

ivanpauno:
> The nav stack defaults several parameters to NOT_SET and then is checking that the use provides a value

I see, yes they seem to be sometimes not passing a default value.

IMHO, the use case we're missing is when you want a "statically typed" parameter with no default value.
In that case, I would expect to declare the parameter passing the parameter type, and if the user didn't provide an override the first "get" throws an exception (or maybe "declare" should throw (?)).

Something like:
```cpp
// throws if an override is not provided, or maybe the first `get()` does and `declare` doesn't throw.
node->declare_parameter(param_name, descriptor, rclcpp::ParameterType::Integer);
```

or

```cpp
// throws if an override is not provided, or maybe the first `get()` does and `declare` doesn't throw.
descriptor.type = ParameterType::Integer;
node->declare_parameter(param_name, descriptor);
```

We would need an `allowed_types` field in the descriptor, or reuse the existing `type` field differently as mentioned [here](https://github.com/ros2/rcl_interfaces/pull/115#issue-516180394).

jacobperron:
I like the `allowed_types` field. We could use a bitmask. This acts as a replacement for the `statically_typed` field, right?

ivanpauno:
> I like the allowed_types field. We could use a bitmask.

I was actually thinking of a single type (i.e. `allowed_type` and not `allowed_types`, sorry for the misspelling), but an `allowed_types` bitmask is also an option.

> This acts as a replacement for the statically_typed field, right?

Yes, we wouldn't need both.

jacobperron:
Maybe I'm wrong, but it seems like we would get a lot more flexibility for little effort in allowing multiple types. I think the API you described would look the same in the single allowed type case. E.g., only allow integers

    node->declare_parameter(param_name, descriptor, rclcpp::ParameterType::Integer);

or allow integers and strings

    node->declare_parameter(param_name, descriptor, rclcpp::ParameterType::Integer | rclcpp::ParameterType::String);

I'm okay if we don't support this, but seems like a nice way to further avoid user-defined parameter callbacks.

ros-discourse:
This pull request has been mentioned on **ROS Discourse**. There might be relevant details there:

https://discourse.ros.org/t/feedback-required-should-parameters-be-able-to-change-of-type/18319/1

ivanpauno:
> I'm okay if we don't support this, but seems like a nice way to further avoid user-defined parameter callbacks.

Yes, I don't mind strongly.
IMO, it looks like a bit of an overkill because I don't think much people is going to use the feature, but it's also not hard to support it.

jacobperron:
It also might help us avoid a future change to the message definition if we decided to support it later.

StefanFabian:
> or allow integers and strings

What's the use case for a parameter that can have such vastly different types?
Can't think of one that wouldn't be bad code design but maybe I'm just not thinking hard enough. It's Friday after all :)

jacobperron:
> What's the use case for a parameter that can have such vastly different types?

Good point. I can't think of any specific cases. And I agree it would be bad design. I think I was under the wrong impression that people were currently using parameters with dynamic types (since that's the current default behavior).

I change my mind and think a single "allowed type" is a good option.

ivanpauno:
> I change my mind and think a single "allowed type" is a good option.

@jacobperron do you have any preference between the currently proposed "static_typing" boolean and "allowed_type" field?
(in both cases you can know the actual type by checking the "type" field)

I would like:

```cpp
node->declare_parameter("an_int", 5);  // parameter_descriptor argument takes default value
```

to declare an integer parameter that cannot change of type.
In the case of an "static_typing" boolean, the default can be true (like the proposed PR here), and you don't have to modify the default parameter descriptor.
In the case of an "allowed_type" field if this behavior is desired, "declare_parameter" would need to overwrite the "allowed_type" field of the provided descriptor and to avoid that behavior we would need an extra flag "dynamic_typing" argument in `declare_parameter`.

Moreover in this case:

```cpp
rcl_interfaces::msg::ParameterDescriptor descriptor;
descriptor.integer_range = /*set integer range to something here*/;
node->declare_parameter("an_int", 5, descriptor);
```

do you need to write `descriptor.allowed_type = rclcpp::PARAMETER_TYPE_INT` or is that implicit?
In the case of `static_typing`, no issue here as it defaults to true.

The only problem is when no default value is provided, that as commented above that seems to be used currently.
I think the options are:

- In case we go ahead with the "static_typing" field, if the user didn't provide a default value we can either:
  - throw an exception
  - set the parameter to "PARAMETER_NOT_SET". "static_typing" starts being enforced when the first actual value is given (taking the type of the first value as the type to enforce).
- In the case of "allowed_type":
  - throw an exception
  - set the parameter to "PARAMETER_NOT_SET". The first value being set must satisfy the allowed type (and after that it cannot be undeclared).

The details are trivial to implement, but I'm not sure what is the preferred way forward ...

jacobperron:
A `static_typing` field (defaulting to true) sounds okay to me.

In the case where no parameter default value is provided, I'd rather not throw an exception since it seems like a valid use-case to me i.e., enforce that the user make a decision on a value instead of letting them use a default. I could be persuaded otherwise though.

ivanpauno:
> i.e., enforce that the user make a decision on a value instead of letting them use a default

I'm not sure what that exactly means.
The only possibility, if not throwing an exception, is to declare the parameter with type "PARAMETER_NOT_SET" and wait until it's set the first time to decide it's type (which is what https://github.com/ros2/rclcpp/pull/1512 does).
But that might be confusing, because it will take the type of the provided override.

ivanpauno:
I'm working on an alternative of the currently opened PRs, in order to experiment what's the best way to modify the behavior without breaking current use cases. I think we can discuss the final details in the review process

jacobperron:
> > i.e., enforce that the user make a decision on a value instead of letting them use a default
>
> I'm not sure what that exactly means.

I mean, I think it's a reasonable use-case to have a "required" parameter. In this case the parameter is not set, requiring the user to set it before it is used. Whether the type is set by the node author or inferred by the first override is another design decision that needs to be made. It would be nice for the node author to be able to specify the allowed type for an unset parameter, but I'm also okay with just inferring the type from the override to keep things simple.

ivanpauno:
Closing in favor of https://github.com/ros2/rcl_interfaces/pull/118.

PR #118, opened 2021-01-21
Enforce static parameter typing

ivanpauno:
Alternative to https://github.com/ros2/rcl_interfaces/pull/115.
See https://github.com/ros2/rclcpp/pull/1522 for details.

### DIFF
--- a/rcl_interfaces/msg/ParameterDescriptor.msg
+++ b/rcl_interfaces/msg/ParameterDescriptor.msg
@@ -21,8 +21,9 @@ string additional_constraints
 # If 'true' then the value cannot change after it has been initialized.
 bool read_only false
 
-# If 'true', the parameter is not allowed to change its type.
-bool static_typing true
+# Type that the parameter is allowed to take.
+# When PARAMETER_NOT_SET (0), the parameter can take any type.
+uint8 allowed_type 0
 
 # If any of the following sequences are not empty, then the constraint inside of
 # them apply to this parameter.
--- a/rcl_interfaces/msg/ParameterDescriptor.msg
+++ b/rcl_interfaces/msg/ParameterDescriptor.msg
@@ -21,8 +21,8 @@ string additional_constraints
 # If 'true' then the value cannot change after it has been initialized.
 bool read_only false
 
-# If 'true', duck typing is allowed.
-bool allow_duck_typing false
+# If 'true', the parameter is not allowed to change its type.
+bool static_typing true
 
 # If any of the following sequences are not empty, then the constraint inside of
 # them apply to this parameter.
--- a/rcl_interfaces/msg/ParameterDescriptor.msg
+++ b/rcl_interfaces/msg/ParameterDescriptor.msg
@@ -21,7 +21,7 @@ string additional_constraints
 # If 'true' then the value cannot change after it has been initialized.
 bool read_only false
 
-# If true, the parameter is allowed to change of type.
+# If true, the parameter is allowed to change type.
 bool dynamic_typing false
 
 # If any of the following sequences are not empty, then the constraint inside of
--- a/rcl_interfaces/msg/ParameterDescriptor.msg
+++ b/rcl_interfaces/msg/ParameterDescriptor.msg
@@ -21,9 +21,8 @@ string additional_constraints
 # If 'true' then the value cannot change after it has been initialized.
 bool read_only false
 
-# Type that the parameter is allowed to take.
-# rcl_interfaces::msg::ParameterType::PARAMETER_DYNAMIC
-uint8 allowed_type 10
+# If true, the parameter is allowed to change of type.
+bool dynamic_typing false
 
 # If any of the following sequences are not empty, then the constraint inside of
 # them apply to this parameter.
--- a/rcl_interfaces/msg/ParameterDescriptor.msg
+++ b/rcl_interfaces/msg/ParameterDescriptor.msg
@@ -22,8 +22,8 @@ string additional_constraints
 bool read_only false
 
 # Type that the parameter is allowed to take.
-# When PARAMETER_NOT_SET (0), the parameter can take any type.
-uint8 allowed_type 0
+# rcl_interfaces::msg::ParameterType::PARAMETER_DYNAMIC
+uint8 allowed_type 10
 
 # If any of the following sequences are not empty, then the constraint inside of
 # them apply to this parameter.
--- a/rcl_interfaces/msg/ParameterDescriptor.msg
+++ b/rcl_interfaces/msg/ParameterDescriptor.msg
@@ -21,6 +21,8 @@ string additional_constraints
 # If 'true' then the value cannot change after it has been initialized.
 bool read_only false
 
+# If 'true', duck typing is allowed.
+bool allow_duck_typing false
 
 # If any of the following sequences are not empty, then the constraint inside of
 # them apply to this parameter.
--- a/rcl_interfaces/msg/ParameterType.msg
+++ b/rcl_interfaces/msg/ParameterType.msg
@@ -2,9 +2,6 @@
 
 # Default value, which implies this is not a valid parameter.
 uint8 PARAMETER_NOT_SET=0
-# Used to specify dynamic typing of a parameter.
-# WARN: if you modify this value, also modify it in rcl_interfaces::msg::ParameterDescriptor
-uint8 PARAMETER_DYNAMIC=10
 
 uint8 PARAMETER_BOOL=1
 uint8 PARAMETER_INTEGER=2
--- a/rcl_interfaces/msg/ParameterType.msg
+++ b/rcl_interfaces/msg/ParameterType.msg
@@ -2,6 +2,9 @@
 
 # Default value, which implies this is not a valid parameter.
 uint8 PARAMETER_NOT_SET=0
+# Used to specify dynamic typing of a parameter.
+# WARN: if you modify this value, also modify it in rcl_interfaces::msg::ParameterDescriptor
+uint8 PARAMETER_DYNAMIC=10
 
 uint8 PARAMETER_BOOL=1
 uint8 PARAMETER_INTEGER=2


jacobperron (approved):

rcl_interfaces/msg/ParameterDescriptor.msg
```diff
@@ -21,6 +21,9 @@ string additional_constraints
 # If 'true' then the value cannot change after it has been initialized.
 bool read_only false
 
+# Type that the parameter is allowed to take.
+# When PARAMETER_NOT_SET (0), the parameter can take any type.
+uint8 allowed_type 0
```
mjeronimo: I'm thinking about the semantics of specifying `allowed_type`. Syntactically, it would be possible to have something like this (pseudo code):

```
type: int
allowed_type: float
```

If so, once a float is assigned, I assume the type would change such that we now have:

```
type: float
allowed_type: float
```
If so, there is kind of a one-way, one-time type conversion allowed and I'm not sure that's what you're going for here. 

Aren't we more interested in whether this is a dynamically-typed parameter or not? In other words:

```
(current) type: int
dynamically_typed: true
```
Then, once assigned another type, such as a float, it would be:

```
(current) type: float
dynamically_typed: true
```
There would be one situation that isn't allowed:

```
type: unknown
dynamically_typed: false
```
That is, one couldn't create a parameter without a known type and also make it statically typed.

jacobperron: This is what was previously proposed in https://github.com/ros2/rcl_interfaces/pull/115 (though using inverse logic calling the member `statically_typed`). The last situation that you described is what motivated this PR: we want to declare a parameter with a static type, but with type "unset". 

Another option, like your example, is to reuse the `type` field to be the "allowed type" in the case of a static type, and the actual type in the dynamic case. But I think mixing the semantics of allowed type and current type could be confusing, hence the introduction of a new `allowed_type` field.

>  Syntactically, it would be possible to have something like this (pseudo code):
>
>       type: int
>       allowed_type: float

While someone could use this message definition like that, IIUC, the implementation in ros2/rclcpp#1522 forbids such a situation.

mjeronimo: Ah, I see, thanks. Still sorting through these PRs to get up to speed. 

I don't see type 'type' field as need to have multiple meanings; it's just the type of the parameter, whether dynamic or not. Now, whether the type can change is a different idea (hence the Boolean), similar to whether the parameter is read-only or not. They are just attributes of the parameter. I think the problem is a statically-typed parameter of which we don't yet know the type. Other than this uninitialized type case, I think the following is clear semantically:

```
type: {int, float, etc.}
static/dynamically_typed: Boolean
read_only: Boolean
etc.
```
To cover this case, it would have to be extended to incorporate an uninitialized/unknown (as we have with UNSET):

```
type: {UNSET, int, float, etc.}
static/dynamically_typed: Boolean
read_only: Boolean
etc.
```
Then, you have to allow the UNSET-to-some-type transition. So far, I'm not seeing the case that can't be represented and has motivated the additional PR (although I'm sure it's there, I'm just not seeing it yet). 

Also, my general philosophy is that handling more at the syntactic level so that invalid combinations can't even be expressed is preferable to a syntax that allows invalid combinations that then must be checked for semantic consistency at run-time; trying to minimize the opportunity for error.

ivanpauno: > I think the problem is a statically-typed parameter of which we don't yet know the type

I think the problem is an uninitialized statically typed parameter, which we already know the desired type.
e.g.:

```cpp
declare_parameter("my_param", PARAMETER_INTEGER);
```

In that case the type is known when declaring, but the problem is to try to [address this comment](https://github.com/ros2/rclcpp/pull/1522#discussion_r562147009).
If we don't want to throw an exception when no override is provided by the user, then we have to declare the parameter with an initial value of PARAMETER_NOT_SET but that it only accepts in the future an integer parameter.
Thus I need both `type` and `allowed_type` in this case, one for being able to know the current type (in this case, uninitialized or integer), and one to know the allowed type (in this case, integer).

If we're happy with throwing an exception at parameter declaration time if no override was provided, then only having a boolean is enough.

(edit) I think I'm repeating more or less what @jacobperron said though.

ivanpauno: > Then, you have to allow the UNSET-to-some-type transition.

The problem with that, is that I would prefer to only allow the UNSET to specific type transition.

> So far, I'm not seeing the case that can't be represented and has motivated the additional PR (although I'm sure it's there, I'm just not seeing it yet).

I'm not completely convinced we should support unset at all, IMO it's easier to just drop it.

mjeronimo: I agree that it would be easier without the UNSET state. It seems to me that parameters should have a well-known initial value (0, 0.0, "", etc.) if the user has not otherwise specified . Also, the descriptions for the declare_parameters methods in the node class say:  "Declare _and initialize_ a parameter with a type." So, I think we either must require an initial value or provide a well-known default.

ivanpauno: Ok, I think that allowing an initial "UNSET" state for statically typed parameters is the only remaining point of discussion, so we only need to settle on that to move forward.
[This other comment](https://github.com/ros2/rclcpp/pull/1522#discussion_r562147009) is exactly related to that.

@jacobperron do you think that keeping the throwing behavior there would be ok?
If someone would want that initial UNSET state, they could support it using a dynamically typed parameter and a custom parameter validation callback.

jacobperron: I suppose keeping the throwing behavior would be okay. The only downside is it's going to cause runtime exceptions for many users switching to Galactic (e.g. the nav2 stack as [pointed out here](https://github.com/ros2/rcl_interfaces/pull/115#issuecomment-736828197)). IIUC, they'll be forced to explicitly provide a default value or make their parameters dynamically typed.

If we decide to disallow statically-typed parameters without a default value, then I suggest we simplify the interface to have a single boolean `dynamically_typed` (or `statically_typed`) field.

ivanpauno: > IIUC, they'll be forced to explicitly provide a default value or make their parameters dynamically typed.

IMO, they wouldn't be forced to that, because they are already throwing!! (see [here](https://github.com/ros-planning/navigation2/blob/11b25a51981eb53db6a6a3601729ae85238d9241/nav2_dwb_controller/dwb_core/src/dwb_local_planner.cpp#L182-L184) and [here](https://github.com/ros-planning/navigation2/blob/11b25a51981eb53db6a6a3601729ae85238d9241/nav2_dwb_controller/dwb_core/src/dwb_local_planner.cpp#L122-L126)).

There are two reasons why you would want to declare a parameter without a default:
- Because a default is meaningless and the user must provide a value.
- For "optional" parameters, which depending in other configurations you will read or not.
  But for those, the right thing to do would be to optionally declare them.

e.g.:
```cpp
auto config = node->declare_parameter("my_config", "variant1");
if ("variant2" == config) {
  int thing_you_only_need_for_config_2 = node->declare_parameter("int_for_config_2", rclcpp::PARAMETER_INTEGER);
  do_something_for_config_2(thing_you_only_need_with_config_2);
}
```

So I think that throwing an exception already handles correctly the main two use cases.

---

The other part of the discussion is if:

```cpp
node->declare_parameter("parameter_name");  // 1
node->declare_parameter("parameter_name", rclcpp::ParameterValue{});  // 2
```

- Should not be valid anymore. e.g.: current PR fails to build for `1` and throws an exception in `2`.
- Should be deprecated in favor of `node->declare_parameter("parameter_name", rclcpp::PARAMETER_DYNAMIC);`
  Problem, deprecating `2` is hard (easy to provide a runtime warning, hard to show a build time deprecation).
- Be still valid, not deprecated.

I think a combination of `1` and `2` is the most sane thing to do.
Deprecate `1`, and make `2` fail at build time, which will break some existing code but not much (hopefully).

jacobperron: Doing what you propose sounds good to me: deprecate `1` and compile error on `2`. Do we still need the new `rcl_interfaces::msg::ParameterType::PARAMETER_DYNAMIC`? I guess we can switch back to using a boolean like in #115.

ivanpauno: yes, if there's agreement with the proposal above we can switch back to a boolean

mjeronimo: > The other part of the discussion is if:
> 
> ```c++
> node->declare_parameter("parameter_name");  // 1
> node->declare_parameter("parameter_name", rclcpp::ParameterValue{});  // 2
> ```
> 
> * Should not be valid anymore. e.g.: current PR fails to build for `1` and throws an exception in `2`.
> * Should be deprecated in favor of `node->declare_parameter("parameter_name", rclcpp::PARAMETER_DYNAMIC);`
>   Problem, deprecating `2` is hard (easy to provide a runtime warning, hard to show a build time deprecation).
> * Be still valid, not deprecated.
> 
> I think a combination of `1` and `2` is the most sane thing to do.
> Deprecate `1`, and make `2` fail at build time, which will break some existing code but not much (hopefully).

I agree with disallowing either of these and catching them both at compile-time, effectively requiring the user to provide an initial value for statically-typed parameters, thereby simplifying the underlying representation (name, type, whether a parameter is dynamically-typed, etc.).

ivanpauno:
@jacobperron @mjeronimo PTAL to this and https://github.com/ros2/rclcpp/pull/1522, most of the feedback was addressed.
Deprecating `node->declare_parameter("name")` is the only pending thing (I think).

ivanpauno:
> Deprecating node->declare_parameter("name") is the only pending thing (I think).

Done that too

rcl_interfaces/msg/ParameterDescriptor.msg
```diff
@@ -21,6 +21,8 @@ string additional_constraints
 # If 'true' then the value cannot change after it has been initialized.
 bool read_only false
 
+# If true, the parameter is allowed to change of type.
```
jacobperron:
```suggestion
# If true, the parameter is allowed to change type.
```

jacobperron (approved):

ivanpauno:
Merging this one before than the others, so I can release and make sure that all the PR checker jobs of downstream PRs are passing.

PR #132, opened 2021-11-30
test_msgs depend on library targets exported with ament_export_targets instead of ament_export_libraries

sloretz:
Part of ament/ament_cmake#365

`test_msgs` has test targets that depend on all library targets created by the rosidl generators. It uses the variable `_AMENT_EXPORT_LIBRARIES` created by `ament_export_libraries()` to get the names of those library targets. This makes it instead depend on `_AMENT_CMAKE_EXPORT_TARGETS` which is created by `ament_export_targets()`.


sloretz:
Actually, this is invalid again. The names in `_AMENT_CMAKE_EXPORT_TARGETS` are the export names, not the target names.

PR #134, opened 2021-12-21
add note that tells read_only parameter can be undeclared.

fujitatomoya:
must be aligned with https://github.com/ros2/rclcpp/pull/1824

note: there is a description about `read_only` parameter in [qos_configurability](http://design.ros2.org/articles/qos_configurability.html), but i do not think that we need to update this section.

Signed-off-by: Tomoya Fujita <Tomoya.Fujita@sony.com>

### DIFF
--- a/rcl_interfaces/msg/ParameterDescriptor.msg
+++ b/rcl_interfaces/msg/ParameterDescriptor.msg
@@ -19,6 +19,7 @@ string description
 string additional_constraints
 
 # If 'true' then the value cannot change after it has been initialized.
+# But parameter still can be undeclared by the node that declared it.
 bool read_only false
 
 # If true, the parameter is allowed to change type.


rcl_interfaces/msg/ParameterDescriptor.msg
```diff
@@ -19,6 +19,7 @@ string description
 string additional_constraints
 
 # If 'true' then the value cannot change after it has been initialized.
+# But parameter still can be undeclared with force flag.
```
alsora: I agree on having a comment here, but the comment should mention that the force flag can be used only by the node declaringthe parameter.

Something like
```
# But parameter still can be undeclared by the node that declared it with the force flag.
```

Actually I'm not sure if I would even mention the force flag.
That's an internal detail that is not relevant for other ROS 2 node (which access the parameter through a parameter client)

So maybe

```
# But parameter still can be undeclared by the node that declared it.
```

fujitatomoya:
closing this aligned with https://github.com/ros2/rclcpp/pull/1824

PR #146, opened 2022-10-11
new package and interfaces for describing other types

wjwwood:
Still in development, but related to "evolving message types" REP-2011: https://github.com/ros-infrastructure/rep/pull/358

There are still TODOs and known issues, but I think can iterate on those after merging this initial MVP. Merging will also unblock others working on related features.


type_description_interfaces/msg/FieldType.msg
```diff
@@ -0,0 +1,149 @@
...
+# A constant for each type supported according to:
+#   http://design.ros2.org/articles/legacy_interface_definition.html
+# and:
+#   http://design.ros2.org/articles/idl_interface_definition.html
+# Order is loosely coupled to the order of appearance in the IDL 4.2 spec:
+#  https://www.omg.org/spec/IDL/4.2
+
+# Layout of constants across the 0-255 decimal values in the uint8:
+#
+# - 000-031: Primitive types, strings, and reserved space for future primitive types
```
wjwwood: Consider groups of 48 for more space for more primitives.

type_description_interfaces/msg/FieldType.msg
```diff
@@ -0,0 +1,149 @@
...
+# - 032-063: Fixed sized array of primitive and string types
+# - 064-095: Bounded Sequences of primitive and string types
+# - 096-127: Unbounded Sequences of primitive and string types
+# - 128-255: Reserved space for future array/sequence-like types
+
+# Nested type defined in other .msg/.idl files.
+uint8 FIELD_TYPE_NESTED_TYPE = 0
```
wjwwood: Use this for "not set".

type_description_interfaces/msg/IndividualTypeDescription.msg
```diff
@@ -0,0 +1,6 @@
+# Represents a single type, without the types it references, if any.
+
+# Name of the type.
+string type_name
+# Fields of the type.
+Field[] fields
```
wjwwood, line 11: Consider adding constants/enums/annotations here, being careful to consider impact to hashing.

wjwwood, line 11: Consider having a padding space for backporting items.

jhurliman:
Is this ready for review, or what components are missing / still in flux?

wjwwood:
@jhurliman There are some TODO's left for me to update. We've been working on other technical parts of the REP lately so this hasn't made any progress lately. However, we will be returning to it before long.

wjwwood:
Also, we took some time to reconsider having our own definition versus re-using something existing like the `dds-xtypes_typeobject.idl` from DDS-XTypes, but we decided to recommend against that and keep our custom format. I have to write up those thoughts into the REP as well.

type_description_interfaces/msg/Field.msg
```diff
@@ -0,0 +1,8 @@
+# Represents a single field in a type.
+
+# Name of the field.
+string name
+# Type of the field, including details about the type like length, nested name, etc.
+FieldType type
+# Literal default value of the field as a string.
```
james-rms: serialized how?

type_description_interfaces/README.md
```diff
@@ -0,0 +1,22 @@
...
+
+This package claims to be in the **Quality Level 1** category, see the [Quality Declaration](QUALITY_DECLARATION.md) for more details.
+
+## TODO
+
+* How to store default values, should store default values? Current storing as string, but will impact hashing.
+* Do we (will we) support fixed sized strings vs. just having bounded strings?
+* Should we pick limits (bounds) for strings in these types, e.g. type names, default values, etc?
+* Should we pick limits (bounds) for sequences in these types, e.g. number of referenced types?
+* How should we handle Services and Actions?
```
emersonknapp: As I'm implementing hashing, I am running into this.

Couple of options I've been thinking of (using Service for example, but applies to Action as well):
1. Only represent Messages, so compatibility/hashing is only on MyServiceRequest, MyServiceResponse, there isn't the higher level matching. This could allow request to be compatible but response to go out of sync, maybe ill advised because of that.
2. Define an `IndividualTypeDescription` for a service, that has just two fields that are referenced types for the implementing messages, something like:

```
type_description:
  type_name: MyService
  fields:
    - name: request
      type:
        type_id: 0 
        nested_type_id: MyServiceRequest
    - name: response
      type:
        type_id: 0
        nested_type_id: MyServiceResponse
referenced_type_descriptions:
  - type_name: MyServiceRequest
    fields:
... etc
```

That second one seems maybe better for real use cases, and it allows us to have a single Type Version Hash at the top level of the interface, so this could possibly be defined in e.g. `idl__struct.h` rather than propagating through e.g. `srv__struct.h` to `msg__struct.h`

type_description_interfaces/README.md
```diff
@@ -0,0 +1,22 @@
...
+* [IndividualTypeDescription](msg/IndividualTypeDescription.msg): Represents a single ROS Message Type by itself
+* [TypeDescription](msg/TypeDescription.msg): Represents a single ROS Message Type along with all referenced types
+
+## Quality Declaration
+
+This package claims to be in the **Quality Level 1** category, see the [Quality Declaration](QUALITY_DECLARATION.md) for more details.
+
+## TODO
+
+* How to store default values, should store default values? Current storing as string, but will impact hashing.
```
emersonknapp: Maybe not relevant anymore since we've decided TypeDescription has to be serialized for hashing, so default values can be skipped in that process.

type_description_interfaces/README.md
```diff
@@ -0,0 +1,22 @@
...
+* Do we (will we) support fixed sized strings vs. just having bounded strings?
```
emersonknapp: rosidl_parser does not have a fixed sized string type, only bounded and unbounded. This type isn't exposed to `.msg`, but I guess a user could specify it in `.idl`?

type_description_interfaces/msg/FieldType.msg
```diff
@@ -0,0 +1,149 @@
...
+uint8 FIELD_TYPE_BYTE_ARRAY= 47
+uint8 FIELD_TYPE_STRING_ARRAY= 48
+uint8 FIELD_TYPE_WSTRING_ARRAY= 49
+uint8 FIELD_TYPE_FIXED_STRING_ARRAY = 50
+uint8 FIELD_TYPE_FIXED_WSTRING_ARRAY = 51
+uint8 FIELD_TYPE_BOUNDED_STRING_ARRAY = 52
+uint8 FIELD_TYPE_BOUNDED_WSTRING_ARRAY = 53
+
+# Bounded Sequence Types
+uint8 FIELD_TYPE_NESTED_TYPE_BOUNDED_SEQUENCE = 64
```
emersonknapp: In translating these values outside this code, I implemented as "first 21, then add a constant based on type". Maybe this would be easier as two fields? The actual payload would be larger but it would reduce constant definitions by 4x.

```
uint8 value_type   # the first 21 values defined here
uint8 sequence_type   # value, array, bounded_sequence, or unbounded_sequence 
```

type_description_interfaces/msg/FieldType.msg
```diff
@@ -0,0 +1,149 @@
...
+# Only used when the type is a fixed or bounded string/wstring, or a array/sequence of those.
+# In the case of a fixed string/wstring, it is the fixed length of the string.
+# In the case of a bounded string/wstring, it is the maximum length of the string.
+# In the case of an array/sequence of fixed string/wstring, it is the fixed length of the strings.
+# In the case of an array/sequence of bounded string/wstring, it is the maximum length of the strings.
+# It is not currently possible to have different string lengths per element in the array/sequences.
+uint64 string_length
+
+# Only used when the type is a nested type or array/sequence of nested types.
+string nested_type_name
```
emersonknapp: should this be bounded string? do type names have a bound in ros2?

emersonknapp:
@clalancette @wjwwood what next steps do you think we need to take to get this landed? perhaps a set of consolidated TODO items will help us get it going. are any further updates needed to REP-2011 before we can move forward with these interfaces, or have we solidified that documented design well enough that it's only implementation details now?

type_description_interfaces/msg/FieldType.msg
```diff
@@ -0,0 +1,149 @@
...
+
+# Byte/Octet Type
+uint8 FIELD_TYPE_BYTE = 15
+
+# String Types
+uint8 FIELD_TYPE_STRING = 16
+uint8 FIELD_TYPE_WSTRING = 17
+
+# Fixed String Types
+uint8 FIELD_TYPE_FIXED_STRING = 18
```
emersonknapp: Note: there is no rosidl_parser.definition for "Fixed String". You can do fixed-size Array of Char, but this precludes "Sequences of Fixed String"s

type_description_interfaces/msg/FieldType.msg
allenh1: I think there's an implicit limit because of the max length of a file name as supported by the OS (~255 - 4 = 251 max), but this does not extend to IDL files AFAIK.

type_description_interfaces/README.md
wjwwood: Yeah, I think so. Or maybe it was that I think we should have a fixed sized string in the future. I think it's ok, but maybe not ideal, that this file can capture more than what you can put in a message. We'll just give up on either end if a message uses this option and we don't yet support it.

type_description_interfaces/msg/FieldType.msg
wjwwood: I was optimizing for size. Also, there's not perfect overlap I think. I'd have to double check that. (cases where there are values that cannot be sequences as well).

wjwwood: I'm not sure optimizing for fewer constants is that important, but I could be wrong.

type_description_interfaces/msg/FieldType.msg
wjwwood: I'm also not sure about this, but yes I could see this either being a fixed (I know we don't support that atm) or bounded sized string.

type_description_interfaces/msg/FieldType.msg
emersonknapp: NOTE: this isn't an issue for translating from IDL->TypeDescription, but it could be a problem for the reverse

type_description_interfaces/msg/FieldType.msg
```diff
@@ -0,0 +1,149 @@
...
+# - 064-095: Bounded Sequences of primitive and string types
+# - 096-127: Unbounded Sequences of primitive and string types
+# - 128-255: Reserved space for future array/sequence-like types
+
+# Nested type defined in other .msg/.idl files.
+uint8 FIELD_TYPE_NESTED_TYPE = 0
+
+# Basic Types
+# Integer Types
+uint8 FIELD_TYPE_INT8 = 1
```
emersonknapp: Note: `rosidl.parser` has the categories of types, `[UN]SIGNED_NONEXPLICIT_INTEGER_TYPES` (containing (`unsigned`) `short`, `long`, `long long`) that are not represented here.
This could be an issue for translating from IDL->TypeDescription, but not the reverse.

type_description_interfaces/msg/Field.msg
wjwwood: I added a comment about this in the docstring, basically stating that for now (we might want to change this later) the value in the string is exactly the string used for the default value in the originating text, whether that was .msg or .idl.

type_description_interfaces/README.md
wjwwood: This seems totally reasonable to me. I don't think we need changes in this pr for this approach, so I'll just give it a 👍  and resolve this conversation. Thanks!

type_description_interfaces/README.md
wjwwood: Right, since you're now ignoring parts of this for the purposes of hashing and we have the default values in already, we can just remove this todo.

type_description_interfaces/msg/FieldType.msg
wjwwood: I think we just leave this and if we encounter this issue (going from TypeDescription.msg to .idl and this option is used), we just throw saying it is not supported.

type_description_interfaces/msg/FieldType.msg
wjwwood: These messages (parts of TypeDescription.msg) are meant to represent what we can use in ROS 2, and since we have a declaration of what the values in msg translate to in idl, I think that we'd use the same matrix here.

Now if someone uses one of these other types in an .idl file, and we want to create a TypeDescription.msg from that, again I would say we use the matrix for converting between .msg and .idl, and if the type used in the .idl is outside of that, we abort saying we don't support it.

type_description_interfaces/msg/IndividualTypeDescription.msg
wjwwood: I added a todo in the README for this.

wjwwood: I added a todo in the README for this.

emersonknapp:
@wjwwood would you mind also rebasing this PR on `rolling`, when checked out it's an issue because of new `service_msgs` package being required by dependents

type_description_interfaces/msg/FieldType.msg
wjwwood: It is bound to 256 bytes, which must include the null terminator, so 255 characters total. See section 9.6.3.2.1 of the DDSI-RTPS spec, table 9.18: https://www.omg.org/spec/DDSI-RTPS/2.5/PDF

I can go ahead and bound this, but like we talked about off-line, bounding just this field doesn't get us much, if we don't also manage to bound the other strings in these messages.

wjwwood: Note, this is the bound for DDS, we could choose something different for ROS 2, but it at least needs to be limited to 255 so we can implement it on our current DDS implementations easily.

wjwwood: I just re-checked and I could not find any documentation about a limit to the type name length in ROS 2's design documents, so that's something we'll have to follow up on in the future, but is out of scope for this pr. I've added notes about it in commit cd119a6.

I could see us further limiting these type name lengths, e.g. so that we can concatenate the type hash with it in some middlewares. On the other hand, those middlewares might not have a 255 cap themselves, allowing us to concatenate the type hash to a ~255 character type name and still not have a problem. These fields are specific the name without the type hash concatenated.

Also, there are still unbounded strings for the field names and the default value, which would need to be bounded to achieve some real benefits here.

wjwwood:
Rebased it on rolling.

wjwwood:
I've moved this out of draft pr mode, and I think it's ready for review.

type_description_interfaces/msg/FieldType.msg
```diff
@@ -0,0 +1,157 @@
...
+uint8 FIELD_TYPE_BYTE_UNBOUNDED_SEQUENCE = 160
+uint8 FIELD_TYPE_STRING_UNBOUNDED_SEQUENCE = 161
+uint8 FIELD_TYPE_WSTRING_UNBOUNDED_SEQUENCE = 162
+uint8 FIELD_TYPE_FIXED_STRING_UNBOUNDED_SEQUENCE = 163
+uint8 FIELD_TYPE_FIXED_WSTRING_UNBOUNDED_SEQUENCE = 164
+uint8 FIELD_TYPE_BOUNDED_STRING_UNBOUNDED_SEQUENCE = 165
+uint8 FIELD_TYPE_BOUNDED_WSTRING_UNBOUNDED_SEQUENCE = 166
+
+# Identifying number for the type of the field, using one of the above constants.
+uint8 type_id
```
methylDragon: Should we explicitly set default as 0? (Even though the uint8 should be default intialized to 0)

type_description_interfaces/msg/FieldType.msg
```diff
@@ -0,0 +1,157 @@
...
+# In the case of a fixed string/wstring, it is the fixed length of the string.
+# In the case of a bounded string/wstring, it is the maximum length of the string.
+# In the case of an array/sequence of fixed string/wstring, it is the fixed length of the strings.
+# In the case of an array/sequence of bounded string/wstring, it is the maximum length of the strings.
+# It is not currently possible to have different string lengths per element in the array/sequences.
+uint64 string_length
+
+# Only used when the type is a nested type or array/sequence of nested types.
+# This is limited to 255 characters.
+# TODO(wjwwood): this 255 character limit was choosen due to this being the limit
```
methylDragon: Nit: 
```suggestion
# TODO(wjwwood): this 255 character limit was chosen due to this being the limit
```

type_description_interfaces/msg/IndividualTypeDescription.msg
```diff
@@ -0,0 +1,11 @@
+# Represents a single type, without the types it references, if any.
+
+# Name of the type.
+# This is limited to 255 characters.
+# TODO(wjwwood): this 255 character limit was choosen due to this being the limit
```
methylDragon:
```suggestion
# TODO(wjwwood): this 255 character limit was chosen due to this being the limit
```

type_description_interfaces/CMakeLists.txt
```diff
@@ -0,0 +1,28 @@
...
+find_package(rosidl_core_generators REQUIRED)
+
+rosidl_generate_interfaces(${PROJECT_NAME}
+  "msg/Field.msg"
+  "msg/FieldType.msg"
+  "msg/IndividualTypeDescription.msg"
+  "msg/TypeDescription.msg"
+  ADD_LINTER_TESTS
+)
+
```
methylDragon, line 25:
```suggestion

if(BUILD_TESTING)
  find_package(ament_lint_auto REQUIRED)
  ament_lint_auto_find_test_dependencies()
endif()
```

type_description_interfaces/package.xml
```diff
@@ -0,0 +1,25 @@
...
+
+  <license>Apache License 2.0</license>
+
+  <buildtool_depend>ament_cmake</buildtool_depend>
+
+  <buildtool_depend>rosidl_core_generators</buildtool_depend>
+
+  <exec_depend>rosidl_core_runtime</exec_depend>
+
+  <test_depend>ament_lint_common</test_depend>
```
methylDragon, line 19:
```suggestion
  <test_depend>ament_lint_auto</test_depend>
  <test_depend>ament_lint_common</test_depend>
```

methylDragon (approved):
Small nits, but other than that I can't think of any fundamental issues with the message definitions

type_description_interfaces/msg/FieldType.msg
allenh1: +1 for this -- I think it's a nice clarification, and I don't think it really has performance implications.

allenh1 (approved):
Looks good to me as is.

type_description_interfaces/msg/FieldType.msg
wjwwood: It would be functionally no different I think, but I can add it for clarity.

type_description_interfaces/CMakeLists.txt
wjwwood: I'll have to double check these changes, because I think these packages might be kind of special since they're low level message packages. I'll double check.

emersonknapp (approved):
LGTM (after fixing typos)

type_description_interfaces/msg/FieldType.msg
```diff
@@ -0,0 +1,157 @@
...
+uint8 FIELD_TYPE_FIXED_WSTRING_UNBOUNDED_SEQUENCE = 164
+uint8 FIELD_TYPE_BOUNDED_STRING_UNBOUNDED_SEQUENCE = 165
+uint8 FIELD_TYPE_BOUNDED_WSTRING_UNBOUNDED_SEQUENCE = 166
+
+# Identifying number for the type of the field, using one of the above constants.
+uint8 type_id
+
+# Only used when the type is an array or a bounded sequence.
+# In the case of an array, this is the fixed length of the array.
+# In the case of a bounded sequence, this is the maximum size of the sequence.
```
allenh1: nit:
```suggestion
# In the case of a bounded sequence, this is the capacity sequence.
```

type_description_interfaces/msg/FieldType.msg
```diff
@@ -0,0 +1,157 @@
...
+# In all other cases this field is unused.
+uint64 length
```
allenh1: I think `length` might be confusing here. `maximum_size` would be a good alternative.

```suggestion
uint64 capacity
```

type_description_interfaces/msg/FieldType.msg
```diff
@@ -0,0 +1,157 @@
...
+
+# Only used when the type is a fixed or bounded string/wstring, or a array/sequence of those.
+# In the case of a fixed string/wstring, it is the fixed length of the string.
+# In the case of a bounded string/wstring, it is the maximum length of the string.
+# In the case of an array/sequence of fixed string/wstring, it is the fixed length of the strings.
+# In the case of an array/sequence of bounded string/wstring, it is the maximum length of the strings.
```
allenh1:
```suggestion
# In the case of an array/sequence of fixed strings/wstrings, it is the fixed length of the strings.
# In the case of an array/sequence of bounded strings/wstrings, it is the maximum length of the strings.
```

type_description_interfaces/msg/FieldType.msg
```diff
@@ -0,0 +1,157 @@
...
+uint8 FIELD_TYPE_BOUNDED_STRING_UNBOUNDED_SEQUENCE = 165
+uint8 FIELD_TYPE_BOUNDED_WSTRING_UNBOUNDED_SEQUENCE = 166
+
+# Identifying number for the type of the field, using one of the above constants.
+uint8 type_id
+
+# Only used when the type is an array or a bounded sequence.
+# In the case of an array, this is the fixed length of the array.
+# In the case of a bounded sequence, this is the maximum size of the sequence.
+# In all other cases this field is unused.
```
allenh1, line 139:
```suggestion
# In all other cases, this field is unused.
```

type_description_interfaces/msg/FieldType.msg
```diff
@@ -0,0 +1,157 @@
...
+uint64 length
+
+# Only used when the type is a fixed or bounded string/wstring, or a array/sequence of those.
+# In the case of a fixed string/wstring, it is the fixed length of the string.
+# In the case of a bounded string/wstring, it is the maximum length of the string.
+# In the case of an array/sequence of fixed string/wstring, it is the fixed length of the strings.
+# In the case of an array/sequence of bounded string/wstring, it is the maximum length of the strings.
+# It is not currently possible to have different string lengths per element in the array/sequences.
```
allenh1: @wjwwood I don't think this statement is true (e.g. `sequence<string>`, and `sequence<string, 255>`).

```suggestion
``` 

Neither of those cases are in the "type is a fixed or bounded string/wstring" category, so I think this statement can just be removed.

type_description_interfaces/msg/FieldType.msg
```diff
@@ -0,0 +1,157 @@
...
+uint64 string_length
```
allenh1: If, at some point in the future, bounded sequences of bounded sequences are supported, it might be better to make this term more generic:

```suggestion
uint64 element_capacity
```

type_description_interfaces/msg/FieldType.msg
```diff
@@ -0,0 +1,157 @@
...
+# - 097-144: Bounded Sequences of primitive and string types
+# - 145-192: Unbounded Sequences of primitive and string types
+# - 193-255: Reserved space for future array/sequence-like types
+
+uint8 FIELD_TYPE_NOT_SET = 0
+
+# Nested type defined in other .msg/.idl files.
+uint8 FIELD_TYPE_NESTED_TYPE = 1
+
+# Basic Types
```
allenh1:
```suggestion
```

There are no other separators like this one, so I think it is superfluous

allenh1 (commented):
@wjwwood on second pass, I have a couple of suggestions.

type_description_interfaces/msg/FieldType.msg
methylDragon:
```suggestion
# In the case of a bounded sequence, this is the capacity of the sequence.
```

type_description_interfaces/msg/FieldType.msg
methylDragon: I'm not sure if max\_size or capacity would give off the impression that a fixed length array could be shortened though, but I don't mind using `capacity` too

type_description_interfaces/msg/FieldType.msg
clalancette: Fixed in 2563dc382581359b57d945b3a8b321599e215030

type_description_interfaces/msg/IndividualTypeDescription.msg
clalancette: Fixed in 2563dc382581359b57d945b3a8b321599e215030

type_description_interfaces/msg/FieldType.msg
clalancette: Fixed in 2563dc382581359b57d945b3a8b321599e215030

type_description_interfaces/CMakeLists.txt
clalancette: It looks like this is fine to do here (a bunch of the other rcl_interfaces packages do it.  So I added it in 2563dc382581359b57d945b3a8b321599e215030

type_description_interfaces/package.xml
clalancette: Added in 2563dc382581359b57d945b3a8b321599e215030

type_description_interfaces/msg/FieldType.msg
clalancette: Fixed in 2563dc382581359b57d945b3a8b321599e215030

type_description_interfaces/msg/FieldType.msg
clalancette: Changed to capacity in 2563dc382581359b57d945b3a8b321599e215030

type_description_interfaces/msg/FieldType.msg
clalancette: I opted to keep this as `string/wstring`, since that is the name of the type.

type_description_interfaces/msg/FieldType.msg
clalancette: This is very much a matter of taste.  While I would also put the comma in, it isn't wrong to not have it.  So I left it as-is.

type_description_interfaces/msg/FieldType.msg
clalancette: I'm not sure I agree.  If we have a sequence of fixed string, for instance, this field represents the fixed length of each of the strings.  It is not possible with this structure to have a sequence of fixed strings, and have some of the strings be of length 10 and some of length 20.  So I think this comment still applies.  @allenh1 were you thinking about something else?

type_description_interfaces/msg/FieldType.msg
clalancette: Agreed, removed in 2563dc382581359b57d945b3a8b321599e215030

type_description_interfaces/msg/FieldType.msg
clalancette: I'm not sure about this one.  @wjwwood could you maybe comment about this?

wjwwood: Since that's not the case now I wouldn't call it element length yet. I think I'm the meantime it would be confusing. Also I'm not sure we'd ever support sequence of sequence.

type_description_interfaces/msg/FieldType.msg
wjwwood: What Chris said is what I was talking about with this line, specifically you cannot have heterogenous sequences and therefore cannot have a sequence with one string fixed to 10 and another fixed to 100 because those are different types.

allenh1: I see. I think we are operating on different definitions for the word "length".

type_description_interfaces/msg/FieldType.msg
```diff
@@ -0,0 +1,156 @@
...
+uint8 FIELD_TYPE_FIXED_STRING_UNBOUNDED_SEQUENCE = 163
+uint8 FIELD_TYPE_FIXED_WSTRING_UNBOUNDED_SEQUENCE = 164
+uint8 FIELD_TYPE_BOUNDED_STRING_UNBOUNDED_SEQUENCE = 165
+uint8 FIELD_TYPE_BOUNDED_WSTRING_UNBOUNDED_SEQUENCE = 166
+
+# Identifying number for the type of the field, using one of the above constants.
+uint8 type_id 0
+
+# Only used when the type is an array or a bounded sequence.
+# In the case of an array, this is the fixed length of the array.
```
clalancette:
```suggestion
# In the case of an array, this is the fixed capacity of the array.
```

type_description_interfaces/msg/FieldType.msg
```diff
@@ -0,0 +1,156 @@
...
+# In the case of a bounded sequence, this is the maximum capacity of the sequence.
+# In all other cases this field is unused.
+uint64 capacity
+
+# Only used when the type is a fixed or bounded string/wstring, or a array/sequence of those.
+# In the case of a fixed string/wstring, it is the fixed length of the string.
+# In the case of a bounded string/wstring, it is the maximum length of the string.
```
clalancette:
```suggestion
# In the case of a bounded string/wstring, it is the maximum capacity of the string.
```

type_description_interfaces/msg/FieldType.msg
```diff
@@ -0,0 +1,156 @@
...
+# In the case of an array/sequence of fixed string/wstring, it is the fixed length of the strings.
+# In the case of an array/sequence of bounded string/wstring, it is the maximum length of the strings.
```
clalancette:
```suggestion
# In the case of an array/sequence of bounded string/wstring, it is the maximum capacity of the strings.
```

type_description_interfaces/msg/FieldType.msg
```diff
@@ -0,0 +1,156 @@
...
+# It is not currently possible to have different string lengths per element in the array/sequence.
```
clalancette:
```suggestion
# It is not currently possible to have different string capacities per element in the array/sequence.
```

type_description_interfaces/msg/FieldType.msg
```diff
@@ -0,0 +1,156 @@
...
+uint64 string_length
```
clalancette:
```suggestion
uint64 string_capacity
```

clalancette:
@allenh1 @wjwwood @methylDragon @emersonknapp @james-rms I've now done the updates to this PR that we discussed last night.  I think this should be ready for final review.  Here is CI:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/18134/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/12680/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/18784/)

clalancette (approved):

wjwwood (commented):
lgtm

allenh1 (approved):
Thanks for the update. Looks good to me.

clalancette:
CI was...very yellow.  Even though I don't think it is the fault of this PR, here is another try:

* Linux [Build Status](http://ci.ros2.org/job/ci_linux/18152/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/12689/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/18810/)

clalancette:
The flake8 failures are known (and fixed elsewhere).  The one additional failure is on Windows, but that is a known flake for right now.  So this all looks good, I'm going to go ahead and merge this in.

PR #147, opened 2022-11-02
lifecycle_msgs: remove non-ASCII chars from field comments

gavanderhoorn:
As per subject.

I'm working on an old platform with an ancient version of GCC which isn't capable of processing files with non-ASCII chars in them.

The unicode apostrophes caused build failures, so 48d66fbb removes them.

### DIFF
--- a/lifecycle_msgs/msg/State.msg
+++ b/lifecycle_msgs/msg/State.msg
@@ -14,7 +14,7 @@ uint8 PRIMARY_STATE_UNCONFIGURED = 1
 # This state represents a node that is not currently performing any processing.
 uint8 PRIMARY_STATE_INACTIVE = 2
 
-# This is the main state of the node’s life cycle. While in this state, the node
+# This is the main state of the node's life cycle. While in this state, the node
 # performs any processing, responds to service requests, reads and processes
 # data, produces output, etc.
 uint8 PRIMARY_STATE_ACTIVE = 3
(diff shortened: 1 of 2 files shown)


gavanderhoorn:
This was never really a problem until we upgraded `rosidl_generator_c` et al. to post https://github.com/ros2/rosidl/pull/593.

clalancette:
I'm torn on this one.

On the one hand, there is no reason that this needs to be Unicode; regular ASCII apostrophe will be just fine, and will fix your use case.

On the other hand, we most definitely should support Unicode inside of comments in `.msg` files.  This was a test of that, though more of an implicit one rather than an explicit one.

I'm inclined to take this PR as-is, and deal with testing Unicode in comments in `.msg` files separately, but I'm not sure.  @ros2/team thoughts?

gavanderhoorn:
> On the other hand, we most definitely should support Unicode inside of comments in `.msg` files. This was a test of that, though more of an implicit one rather than an explicit one.

Didn't https://github.com/ros2/rcl_interfaces/pull/80 already add those?

---

Edit: oh, hm, you wrote "in comments". #80 does not test comments.

My vote would obviously be: implicit tests are no tests, and fix those missing tests in separate PR(s) ;)

sloretz (approved):
Changes LGTM!

My vote would be opening an issue on https://github.com/ros2/rosidl/issues about adding tests for non-ascii characters in comments in message files.

clalancette:
> My vote would be opening an issue on https://github.com/ros2/rosidl/issues about adding tests for non-ascii characters in comments in message files.

That works for me.  @gavanderhoorn Would you mind opening up that issue?  Then we can run CI on this.

gavanderhoorn:
See https://github.com/ros2/rosidl/issues/713.

clalancette (approved):

clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/17578/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/12118/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/18144/)

gavanderhoorn:
Thanks for merging.

Would this be back-portable to Humble as well?

clalancette:
Yeah, that seems reasonable.

@Mergifyio backport humble

clalancette:
@Mergifyio backport humble

PR #153, opened 2023-03-02
Add GetTypeDescription.srv (rep2011)

emersonknapp:
Part of https://github.com/ros2/ros2/issues/1159

Adds service interface for retrieving type descriptions from nodes

### DIFF
--- a/type_description_interfaces/msg/TypeSource.msg
+++ b/type_description_interfaces/msg/TypeSource.msg
@@ -4,9 +4,13 @@
 string type_name
 
 # The type of the original source file, typically matching the file extension.
-# Well-known encodings: "idl", "msg", "srv", "action", "dynamic".
+# Well-known encodings: "idl", "msg", "srv", "action", "dynamic", "implicit".
+# "dynamic" specifies a type created programmatically by a user, thus having no source.
+# "implicit" specifies a type created automatically as a subtype of a
+# complex type (service or action) - such as the request message for a service.
+# Implicit types will have no contents, the full source will be available on the parent srv/action.
 string encoding
 
 # Dumped contents of the interface definition source file.
-# If this was a type created programmatically (encoding "dynamic"), this field will be empty.
+# If `encoding` is "dynamic" or "implicit", this field will be empty.
 string raw_file_contents
--- a/type_description_interfaces/msg/TypeSource.msg
+++ b/type_description_interfaces/msg/TypeSource.msg
@@ -1,0 +1,16 @@
+# Represents the original source of a ROS 2 interface definition.
+
+uint8 ENCODING_NOT_SET = 0
+uint8 ENCODING_MSG = 1
+uint8 ENCODING_SRV = 2
+uint8 ENCODING_ACTION = 3
+uint8 ENCODING_IDL = 4
+
+# ROS interface type name, in PACKAGE/NAMESPACE/TYPENAME format
+string type_name
+
+# The type of the original source file, from the above constants
+uint8 encoding
+
+# Dumped contents of the interface definition source file
+string raw_file_contents
--- a/type_description_interfaces/msg/TypeSource.msg
+++ b/type_description_interfaces/msg/TypeSource.msg
@@ -1,16 +1,12 @@
 # Represents the original source of a ROS 2 interface definition.
 
-uint8 ENCODING_NOT_SET = 0
-uint8 ENCODING_MSG = 1
-uint8 ENCODING_SRV = 2
-uint8 ENCODING_ACTION = 3
-uint8 ENCODING_IDL = 4
-
-# ROS interface type name, in PACKAGE/NAMESPACE/TYPENAME format
+# ROS interface type name, in PACKAGE/NAMESPACE/TYPENAME format.
 string type_name
 
-# The type of the original source file, from the above constants
-uint8 encoding
+# The type of the original source file, typically matching the file extension.
+# Well-known encodings: "idl", "msg", "srv", "action", "dynamic".
+string encoding
 
-# Dumped contents of the interface definition source file
+# Dumped contents of the interface definition source file.
+# If this was a type created programmatically (encoding "dynamic"), this field will be empty.
 string raw_file_contents
--- a/type_description_interfaces/package.xml
+++ b/type_description_interfaces/package.xml
@@ -12,6 +12,7 @@
   <buildtool_depend>ament_cmake</buildtool_depend>
 
   <buildtool_depend>rosidl_core_generators</buildtool_depend>
+  <depend>service_msgs</depend>
 
   <exec_depend>rosidl_core_runtime</exec_depend>
 
--- a/type_description_interfaces/package.xml
+++ b/type_description_interfaces/package.xml
@@ -12,6 +12,7 @@
   <buildtool_depend>ament_cmake</buildtool_depend>
 
   <buildtool_depend>rosidl_core_generators</buildtool_depend>
+
   <depend>service_msgs</depend>
 
   <exec_depend>rosidl_core_runtime</exec_depend>
--- a/type_description_interfaces/srv/GetTypeDescription.srv
+++ b/type_description_interfaces/srv/GetTypeDescription.srv
@@ -17,7 +17,10 @@ string failure_reason
 # The parsed type description which can be used programmatically.
 TypeDescription type_description
 
-# List of sources, including all comments and whitespace.
+# A list containing the interface definition source text of the requested type,
+# plus all types it recursively depends on.
+# Each source text is a copy of the original contents of the
+# .msg, .srv, .action, .idl, or other file if it exists, including comments and whitespace.
 # Sources can be matched with IndividualTypeDescriptions by their `type_name`.
 # The `encoding` field of each entry informs how to interpret its contents.
 TypeSource[] type_sources
--- a/type_description_interfaces/srv/GetTypeDescription.srv
+++ b/type_description_interfaces/srv/GetTypeDescription.srv
@@ -1,4 +1,7 @@
+# ROS interface type name, in PACKAGE/NAMESPACE/TYPENAME format
 string type_name
+
+# REP-2011 RIHS hash string, which looks like RIHSxx_hashvalue
 string type_version_hash
 ---
 # True if the type description information is available and populated in the response
@@ -8,9 +11,14 @@ string failure_reason
 
 # The idl or msg file name
 string type_description_raw_file_name
+
 # The idl or msg file, with comments and whitespace
 # The file extension and/or the contents can be used to determine the format
+# This contains all recursive raw files, each separated by a line of 80 '=' characters
+# The order will be the same as the TypeDescription -
+# which is top level interface then referenced type descriptions in alphabetical order by type_name
 string type_description_raw
+
 # The parsed type description which can be used programmatically
 TypeDescription type_description
 
--- a/type_description_interfaces/srv/GetTypeDescription.srv
+++ b/type_description_interfaces/srv/GetTypeDescription.srv
@@ -2,7 +2,7 @@
 string type_name
 
 # REP-2011 RIHS hash string, which looks like RIHSxx_hashvalue
-string type_version_hash
+string type_hash
 
 # Whether to return the original idl or msg file(s) in the response, if available
 bool include_type_sources true
--- a/type_description_interfaces/srv/GetTypeDescription.srv
+++ b/type_description_interfaces/srv/GetTypeDescription.srv
@@ -4,37 +4,30 @@ string type_name
 # REP-2011 RIHS hash string, which looks like RIHSxx_hashvalue
 string type_version_hash
 
-# Whether to return the raw idl or msg file(s) in the response, if available
-bool include_raw_description true
+# Whether to return the original idl or msg file(s) in the response, if available
+bool include_type_sources true
 ---
-# include_raw_description requested, and provided successfully
-uint8 RAW_DESCRIPTION_REQUESTED_PROVIDED = 0
-# include_raw_description disabled, so type_description_raw was left empty
-uint8 RAW_DESCRIPTION_NOT_REQUESTED_NOT_PROVIDED = 1
-# include_raw_description requested, but no raw version available (perhaps for dynamically created types)
-uint8 RAW_DESCRIPTION_REQUESTED_NOT_AVAILABLE = 2
+# sources requested, and provided successfully
+uint8 SOURCES_REQUESTED_PROVIDED = 0
+# sources disabled, so type_sources was left empty
+uint8 SOURCES_NOT_REQUESTED_NOT_PROVIDED = 1
+# sources requested, but no source available (perhaps for dynamically created types)
+uint8 SOURCES_REQUESTED_NOT_AVAILABLE = 2
 
 # True if the type description information is available and populated in the response
 # If false, all other fields except `reason` are considered undefined
 bool successful
 # If `successful` is false, contains a reason for failure.
-# On success, may contain a reason for missing type_description_raw.
+# On success, may contain a reason for missing sources.
 string reason
 
-# Contains a value from RAW_DESCRIPTION_ constants, explaining whether the
-# `type_description_raw_file_name` and `type_description_raw` fields contain valid data.
-uint8 type_description_raw_result
+# Contains a value from above constants, explaining whether the
+# `type_sources` fields contain valid data.
+uint8 type_sources_result
 
-# The idl or msg file name
-# Empty if not available, which will be noted in `reason`
-string type_description_raw_file_name
-
-# The idl or msg file, with comments and whitespace
-# The file extension and/or the contents can be used to determine the format
-# This contains all recursive raw files, each separated by a line of 80 '=' characters
-# The order will be the same as the TypeDescription -
-# which is top level interface, then referenced type descriptions in alphabetical order by type_name
-string type_description_raw
+# List of idl or msg/srv/action source files, including all comments and whitespace.
+# Sources can be matched with IndividualTypeDescriptions by their type_name
+TypeSource[] type_sources
 
 # The parsed type description which can be used programmatically
 TypeDescription type_description
--- a/type_description_interfaces/srv/GetTypeDescription.srv
+++ b/type_description_interfaces/srv/GetTypeDescription.srv
@@ -11,17 +11,18 @@ bool include_raw_description true
 uint8 RAW_DESCRIPTION_REQUESTED_PROVIDED = 0
 # include_raw_description disabled, so type_description_raw was left empty
 uint8 RAW_DESCRIPTION_NOT_REQUESTED_NOT_PROVIDED = 1
-# include_raw_description requested, but no raw version available (perhaps for )
+# include_raw_description requested, but no raw version available (perhaps for dynamically created types)
 uint8 RAW_DESCRIPTION_REQUESTED_NOT_AVAILABLE = 2
 
 # True if the type description information is available and populated in the response
+# If false, all other fields except `reason` are considered undefined
 bool successful
 # If `successful` is false, contains a reason for failure.
 # On success, may contain a reason for missing type_description_raw.
 string reason
 
-# Contains a value from RAW_DESCRIPTION_ constants,
-# explaining the value of `type_description_raw_file_name` and  `type_description_raw`
+# Contains a value from RAW_DESCRIPTION_ constants, explaining whether the
+# `type_description_raw_file_name` and `type_description_raw` fields contain valid data.
 uint8 type_description_raw_result
 
 # The idl or msg file name
--- a/type_description_interfaces/srv/GetTypeDescription.srv
+++ b/type_description_interfaces/srv/GetTypeDescription.srv
@@ -3,25 +3,40 @@ string type_name
 
 # REP-2011 RIHS hash string, which looks like RIHSxx_hashvalue
 string type_version_hash
+
+# Whether to return the raw idl or msg file(s) in the response, if available
+bool include_raw_description true
 ---
+# include_raw_description requested, and provided successfully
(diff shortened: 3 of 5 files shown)


emersonknapp:
@allenh1 @methylDragon @clalancette @wjwwood requesting review, this has no dependencies

type_description_interfaces/package.xml
```diff
@@ -12,6 +12,7 @@
   <buildtool_depend>ament_cmake</buildtool_depend>
 
   <buildtool_depend>rosidl_core_generators</buildtool_depend>
+  <depend>service_msgs</depend>
 
   <exec_depend>rosidl_core_runtime</exec_depend>
```
wjwwood: nitpick: if we're going to have spaces between groups of dependencies, makes sense to have them everywhere (and `<depend>` isn't obviously with one or the other).

```suggestion
  <buildtool_depend>rosidl_core_generators</buildtool_depend>

  <depend>service_msgs</depend>

  <exec_depend>rosidl_core_runtime</exec_depend>

```

type_description_interfaces/srv/GetTypeDescription.srv
```diff
@@ -0,0 +1,19 @@
+string type_name
+string type_version_hash
```
wjwwood: We should document here how these strings are expected to be structured, e.g. `type_name` should be `x/y/z` or maybe `type_version_hash` is expected to be `RIHS...` whatever.

type_description_interfaces/srv/GetTypeDescription.srv
```diff
@@ -0,0 +1,19 @@
...
+# True if the type description information is available and populated in the response
+bool successful
+# Empty if 'successful' was true, otherwise contains details on why it failed
+string failure_reason
+
+# The idl or msg file name
+string type_description_raw_file_name
+# The idl or msg file, with comments and whitespace
+# The file extension and/or the contents can be used to determine the format
+string type_description_raw
```
wjwwood: I think this was brought up in the REP but maybe never resolved, but this should contain the raw descriptions recursively and therefore we need to have a format for storing them.

We should probably figure that out and document it here. Or tentatively we could have a TODO for that and fix it up later, because presumably we can just update the comment without changing the actual interface of this service.

type_description_interfaces/srv/GetTypeDescription.srv
emersonknapp: Added comments describing expected format

type_description_interfaces/srv/GetTypeDescription.srv
emersonknapp: I've added a comment with my proposal for the format. It follows a ROS-classic precedent that I've seen somewhere, which is currently used in MCAP schema records. That is, use a delimiter line of 80 * '=' between individual file contents:

```
file_contents
================================================================================
next_file_contents
================================================================================
next_file_contents
```

Related, do you think we should we make the raw output optional, specified by a `bool` in the request?

type_description_interfaces/srv/GetTypeDescription.srv
```diff
@@ -0,0 +1,27 @@
...
+
+# The idl or msg file name
+string type_description_raw_file_name
+
+# The idl or msg file, with comments and whitespace
+# The file extension and/or the contents can be used to determine the format
+# This contains all recursive raw files, each separated by a line of 80 '=' characters
+# The order will be the same as the TypeDescription -
+# which is top level interface then referenced type descriptions in alphabetical order by type_name
+string type_description_raw
```
allenh1: I'm not so sure I like shoving all of this in the response. I would prefer this service return _just_ the type description, and that we make the rest of the information available through other calls.

What does this contain when the dynamic types API is used to construct a type on the fly?

emersonknapp: What about switchable via a `bool` in the request? That way we don't have to define two interfaces, but we can switch on or off the overhead of including all the raw text.

The raw text would have to be empty for a dynamic type, I assume. There is no "source". Maybe there's a bool in the response stating whether it's a system-loaded type or a dynamic one?

wjwwood: We discussed this in the REP and decided to not have a flag to control this behavior. We can reevaluate that decision, but we did discuss/decide on it already.

type_description_interfaces/srv/GetTypeDescription.srv
wjwwood: I think that works.

As for the option, we discussed that in the REP and decided against it, but we can look at that again if we think we need it.

type_description_interfaces/srv/GetTypeDescription.srv
emersonknapp: What about the question of, what do we put in `type_description_raw` for dynamic types?

clalancette: From the in-person conversation:
1.  We decided to add in a boolean to control whether we return the description_raw at all.
2. Add in an enum to describe what is in the raw field:
  a. DESCRIPTION_REQUESTED_PROVIDED
  b. DESCRIPTION_NOT_REQUESTED_NOT_PROVIDED
  c. DESCRIPTION_REQUESTED_NOT_AVAILABLE

type_description_interfaces/srv/GetTypeDescription.srv
```diff
@@ -0,0 +1,27 @@
...
+
+# The parsed type description which can be used programmatically
+TypeDescription type_description
+
+# Key-value pairs of extra information.
+string[] extra_information_keys
+string[] extra_information_values
```
clalancette: Let's make a `KeyValue[] extra_information` instead of separate key and value arrays.

type_description_interfaces/srv/GetTypeDescription.srv
```diff
@@ -0,0 +1,27 @@
+# ROS interface type name, in PACKAGE/NAMESPACE/TYPENAME format
+string type_name
+
+# REP-2011 RIHS hash string, which looks like RIHSxx_hashvalue
+string type_version_hash
+---
+# True if the type description information is available and populated in the response
+bool successful
+# Empty if 'successful' was true, otherwise contains details on why it failed
```
clalancette: Change this to allow for reasons even if successful.

type_description_interfaces/srv/GetTypeDescription.srv
```diff
@@ -0,0 +1,27 @@
...
+string failure_reason
```
clalancette, line 15: Rename this to `reason` instead of `failure_reason`.

clalancette (changes requested):
Changes from in-person discussion.

emersonknapp:
funny how it doesn't let you re-request all reviews, it removes everyone else if you re-request one person. anyway @allenh1 @clalancette @wjwwood, updated per discussion, happy to get feedback on the language in there, I tried to make it as clear as I could

type_description_interfaces/srv/GetTypeDescription.srv
```diff
@@ -0,0 +1,42 @@
...
+string type_version_hash
+
+# Whether to return the raw idl or msg file(s) in the response, if available
+bool include_raw_description true
+---
+# include_raw_description requested, and provided successfully
+uint8 RAW_DESCRIPTION_REQUESTED_PROVIDED = 0
+# include_raw_description disabled, so type_description_raw was left empty
+uint8 RAW_DESCRIPTION_NOT_REQUESTED_NOT_PROVIDED = 1
+# include_raw_description requested, but no raw version available (perhaps for )
```
wjwwood: Looks like an unfinished thought here.

methylDragon:
I was looking at the REP again and realized that we didn't discuss in detail how we would be including information about the serialization library used to generate the buffer on the wire. I was wondering if it makes sense to have a "serialization library name" and/or version field in the service response for the publishers to state what form of buffer they are publishing, or if that sort of information should be in the implementation specific key-value pairs instead...

I understand from [the discussion on the REP](https://github.com/ros-infrastructure/rep/pull/358#discussion_r1065413893) that we're not intending to support cross protocol communications, but information like serialization library name in my opinion is universalisable enough that putting it in an implementation specific key-value pair is probably not the right call, I think?

This might be out of scope, but I'm thinking of a use case where a user might send a protobuf byte buffer (as a bytestring) through FastRTPS, then comes out the other end and has to be interpreted. (i.e. the middleware implementation is treated as the same, but not the serialization.) Or perhaps more saliently, the bytestrings are stored in a rosbag and need to be interpreted.

wjwwood:
> This might be out of scope, but I'm thinking of a use case where a user might send a protobuf byte buffer (as a bytestring) through FastRTPS, then comes out the other end and has to be interpreted. (i.e. the middleware implementation is treated as the same, but not the serialization.) Or perhaps more saliently, the bytestrings are stored in a rosbag and need to be interpreted.

This must be handled on the transport layer in my opinion. We could include this information here, but I don't think it's actionable information. For example, if you are subscribing with fastrtps that uses CDR, and they are publishing fastrtps that is using protobuf, then they should never match and exchange data. The fact that this information might be reflected in the key-value pairs would only possibly be useful for debugging?

Put a different way, the key-value pairs should only be used to exchange information that could not be gathered on the receiving end, and I think the serialization type is either the same (and therefore you can get it on the subscription side) or is different and they should never exchange data.

However, something like library version might be interesting. In that case it might be that they match and communicate (say Humble to Rolling or something), but the libraries being used might have different versions, and this could potentially be useful information and information that the subscription side doesn't have.

type_description_interfaces/srv/GetTypeDescription.srv
```diff
@@ -0,0 +1,42 @@
...
+uint8 RAW_DESCRIPTION_REQUESTED_NOT_AVAILABLE = 2
+
+# True if the type description information is available and populated in the response
```
clalancette: Just to be extra pedantic here, we might want to say that if `successful` is False, all fields below are considered undefined except for `reason`.

type_description_interfaces/srv/GetTypeDescription.srv
```diff
@@ -0,0 +1,42 @@
...
+bool successful
+# If `successful` is false, contains a reason for failure.
+# On success, may contain a reason for missing type_description_raw.
+string reason
+
+# Contains a value from RAW_DESCRIPTION_ constants,
+# explaining the value of `type_description_raw_file_name` and  `type_description_raw`
```
clalancette:
```suggestion
# explaining whether `type_description_raw_file_name` and  `type_description_raw` contain valid data.
```

(or something)

methylDragon:
> > This might be out of scope, but I'm thinking of a use case where a user might send a protobuf byte buffer (as a bytestring) through FastRTPS, then comes out the other end and has to be interpreted. (i.e. the middleware implementation is treated as the same, but not the serialization.) Or perhaps more saliently, the bytestrings are stored in a rosbag and need to be interpreted.
> 
> This must be handled on the transport layer in my opinion. We could include this information here, but I don't think it's actionable information. For example, if you a subscribing with fastrtps that uses CDR, and they are publishing fastrtps that is using protobuf, then they should never match and exchange data. The fact that this information might be reflected in the key-value pairs would only possibly be useful for debugging?
> 
> Put a different way, the key-value pairs should only be used to exchange information that could not be gathered on the receiving end, and I think the serialization type is either the same (and therefore you can get it on the subscription side) or is different and they should never exchange data.
> 
> However, something like library version might be interesting. In that case it might be that they match and communicate (say Humble to Rolling or something), but the libraries being used might have different versions, and this could potentially be useful information and information that the subscription side doesn't have.

I'm still a little unclear on this; I thought it was possible for FastRTPS to publish and subscribe to an opaque bytestream, and if that were the case, that it would be possible to shove in any bytestream from any serialization library and send it on the wire (as long as both the pub and sub are using FastRTPS as the middleware.)

So in that case, wouldn't there be a case where the middlewares match, the endpoints match, but the buffers themselves can't get interpreted (unless the correct serialization library is used to deserialize the messages?)

---

On the flipside, I can see how getting the `GetTypeDescription` service to work in the different-serialization-scenario described above to be pretty tough... I suppose then the only valid use-case would be if both sides are natively using the same middleware and serialization library, but for some reason the user decided to publish dynamic data with a **different** serialization library.

wjwwood:
> I'm still a little unclear on this; I thought it was possible for FastRTPS to publish and subscribe to an opaque bytestream, and if that were the case, that it would be possible to shove in any bytestream from any serialization library and send it on the wire (as long as both the pub and sub are using FastRTPS as the middleware.)
> 
> So in that case, wouldn't there be a case where the middlewares match, the endpoints match, but the buffers themselves can't get interpreted (unless the correct serialization library is used to deserialize the messages?)

In this case, I would argue that a) the serialization type of the content in the octet sequence should be communicated in the type name, e.g. `SerializedProtobuf.msg`, or some other external way, and that b) the middleware does not and should not know this information, and so I don't know how it would be added to these key-value pairs. I suppose anyone could put anything they wanted in the key-value pairs with some changes to `rcl` (or if they had a non-`rcl` based client library), but this is still not the job of the middleware.

wjwwood:
>  but for some reason the user decided to publish dynamic data with a different serialization library.

That just cannot work, I think.

wjwwood:
> I'm still a little unclear on this; I thought it was possible for FastRTPS to publish and subscribe to an opaque bytestream, and if that were the case, that it would be possible to shove in any bytestream from any serialization library and send it on the wire (as long as both the pub and sub are using FastRTPS as the middleware.)

Sorry, I think I misunderstood your original point, it's not a message that contains bytes, but that you're using the "take serialized" methods from Fast-RTPS so it presumably doesn't have to deserialize it. If it tried that would fail because Fast-RTPS is expecting CDR and the data is something else.

Again, in this situation, that cannot happen. Deserializing or not, Fast-RTPS cannot match an endpoint that uses a different serialization scheme. I do not think that's current possible, and I don't think it should be allowed either.

emersonknapp:
Gist: https://gist.githubusercontent.com/emersonknapp/4859c688aec691dc8d1f95fc9478e54f/raw/0d17f281cd87c92cf68378d663c4023280970d51/ros2.repos
BUILD args:  --packages-above-and-dependencies type_description_interfaces
TEST args:  --packages-above type_description_interfaces
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/11644
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/18277/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/12807/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/18947/)

type_description_interfaces/srv/GetTypeDescription.srv
```diff
@@ -0,0 +1,43 @@
...
+# The idl or msg file name
+# Empty if not available, which will be noted in `reason`
+string type_description_raw_file_name
+
+# The idl or msg file, with comments and whitespace
+# The file extension and/or the contents can be used to determine the format
+# This contains all recursive raw files, each separated by a line of 80 '=' characters
+# The order will be the same as the TypeDescription -
+# which is top level interface, then referenced type descriptions in alphabetical order by type_name
+string type_description_raw
```
clalancette: After discussion in the REP-2011 meeting:

The single type description filename is not enough on its own; we actually need the file names of all types.

What we decided to do here was to instead create a structure that contains three fields:

1. The file name
2. The "encoding" (msg or IDL)
3. The description

And then in this message we have an array of that structure.

emersonknapp: OK - new structure added, and some language changed to call it "sources" rather than "raw descriptions" because that makes a lot more sense given what it is

emersonknapp:
@wjwwood @james-rms @allenh1 updated, changed the language around a little bit

type_description_interfaces/msg/TypeSource.msg
```diff
@@ -0,0 +1,16 @@
...
+uint8 ENCODING_MSG = 1
+uint8 ENCODING_SRV = 2
+uint8 ENCODING_ACTION = 3
+uint8 ENCODING_IDL = 4
+
+# ROS interface type name, in PACKAGE/NAMESPACE/TYPENAME format
+string type_name
+
+# The type of the original source file, from the above constants
+uint8 encoding
```
wjwwood: I'm not sure about the use of an enum for this. Makes it hard to expand in the future. I was thinking we'd use a string, and for now default to using the file extension as the identifiers.

Imagine the process for using protobuf here.

type_description_interfaces/msg/TypeSource.msg
```diff
@@ -0,0 +1,16 @@
...
+
+# Dumped contents of the interface definition source file
+string raw_file_contents
```
wjwwood: What happens if this type was dynamically created, is this empty? What is the encoding set to?

I think we need more doc strings here describing the various scenarios.

type_description_interfaces/srv/GetTypeDescription.srv
```diff
@@ -0,0 +1,36 @@
...
+# True if the type description information is available and populated in the response
+# If false, all other fields except `reason` are considered undefined
+bool successful
+# If `successful` is false, contains a reason for failure.
+# On success, may contain a reason for missing sources.
+string reason
+
+# Contains a value from above constants, explaining whether the
+# `type_sources` fields contain valid data.
+uint8 type_sources_result
```
wjwwood: Now I'm wondering if we need this status field for each source, e.g. if you request this for a dynamically created type, the dynamically created type will have a name, but no sources, but it can depend on static types that do have sources, and so you can have partially filled `type_sources`.

Similarly, now you might have different reasons for each individual type not having it's sources even when requested, so reusing the `reason` string above may no longer make sense.

type_description_interfaces/msg/TypeSource.msg
emersonknapp: Yes, I see your point - changing to a string and listing the well-known values

type_description_interfaces/msg/TypeSource.msg
emersonknapp: I'm adding a note for "well-known encoding" `"dynamic"`, which indicates a type created programmatically, and a note on this field that for `"dynamic"` it will be empty.

type_description_interfaces/srv/GetTypeDescription.srv
emersonknapp: Yes, I think you're right about this. I'm inclined to say now that we return a `TypeSource` entry for every type, and if it  was generated from a source then it contains sources, if it's dynamically created the entry is `encoding: "dynamic"`. There is no `success: true` case where sources are not provided - so there is no need for an enum and an overloading of `reason`

allenh1 (approved):
@emersonknapp thanks! LGTM

type_description_interfaces/srv/GetTypeDescription.srv
achim-k: I think something went wrong here, this became `failure_reason` again (1db947856ece4ba136671fc5136f2af2ef7f1132)

type_description_interfaces/srv/GetTypeDescription.srv
```diff
@@ -0,0 +1,26 @@
...
+# If `successful` is false, contains a reason for failure.
+# If `successful` is true, this is left empty.
+string failure_reason
+
+# The parsed type description which can be used programmatically.
+TypeDescription type_description
+
+# List of sources, including all comments and whitespace.
+# Sources can be matched with IndividualTypeDescriptions by their `type_name`.
+# The `encoding` field of each entry informs how to interpret its contents.
```
clalancette: I think if someone were to look at this cold, it would be hard for them to understand a) what a "source" is, and b) that this is a list of the type that was requested, and all of the recursive types it depends on.  For that reason, I think we should update this comment to something like:

```suggestion
# A list containing the interface definition source text of the originally requested type,
# along with all of the types it depends on (recursively).
# Each source text is a copy of the original contents of the
# `.msg`, `.srv`, `.action`, or `.idl` file, including comments and whitespace.
# Sources can be matched with IndividualTypeDescriptions in the `type_description` by
# their `type_name`.
# The `encoding` field of each entry informs how to interpret its contents.
```

clalancette (approved):
I've left one suggestion on how to clarify the description of `type_sources`.  Otherwise, I really like how this turned out; this is much nicer.  I'm going to approve and assume you do something to update that comment (event if you don't take exactly what I suggested).

type_description_interfaces/srv/GetTypeDescription.srv
wjwwood: I think that might have been intentional since it's no longer multi purpose? I'm not sure I haven't re-reviewed this yet.

emersonknapp: Correct, changed back on purpose, see https://github.com/ros2/rcl_interfaces/pull/153#discussion_r1144661471

type_description_interfaces/srv/GetTypeDescription.srv
emersonknapp: Updated to approximately this language, plus more detail in `TypeSource.msg`. Added a new "well-known encoding" - "implicit" for subtypes of complex types (services/actions), where the source will be available at the `.srv`/`.action` level.

emersonknapp:
Gist: https://gist.githubusercontent.com/emersonknapp/3de13aa51c4d4c48a63baaf3a0ec801c/raw/0d17f281cd87c92cf68378d663c4023280970d51/ros2.repos
BUILD args:  --packages-above-and-dependencies type_description_interfaces
TEST args:  --packages-above type_description_interfaces
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/11659
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/18296/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/12822/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/18965/)

PR #155, opened 2023-03-10
Runtime Interface Reflection: type_description_interfaces_utils

methylDragon:
This PR is part of the runtime interface reflection subscription feature of REP-2011: https://github.com/ros2/ros2/issues/1374

# Description
This PR adds a utilities library for manipulating and constructing the C structs generated by `type_description_interfaces`, including:
- Handy print functions
- Hash map construction
- Sorting and pruning of referenced types (to prep them for hashing)
- Checking of type validity
- Appending fields and referenced type descriptions
- Creating TypeDescriptions from referenced type descriptions (description subsetting)

I also pre-generate the `type_description_interfaces` sources and headers to create a new package (`type_description_interfaces_static`) to remove a dependency cycle downstream. (That's why there are so many lines.)

~~I haven't implemented tests yet, so some memory management issues/segfaults are expected. I'm working on it, but I am wary of spending too much time getting 100% coverage because we're trying to meet the deadline for feature freeze...~~

I've implemented a bunch of tests and I'm fairly confident that most of the core functionalities work as expected now, so I'm going to mark the PR as ready to review.

[Asan](https://github.com/colcon/colcon-sanitizer-reports) runs green too!
```shell
clear && \
colcon build --packages-select type_description_interfaces_utils \
  --build-base=build-asan --install-base=install-asan \
  --cmake-args -DOSRF_TESTING_TOOLS_CPP_DISABLE_MEMORY_TOOLS=ON \
               -DINSTALL_EXAMPLES=OFF -DSECURITY=ON --no-warn-unused-cli \
               -DCMAKE_BUILD_TYPE=Debug \
  --mixin asan-gcc && \
colcon test --build-base=build-asan --install-base=install-asan \
  --event-handlers sanitizer_report+ --packages-select type_description_interfaces_utils
```

There are some less important tests waiting to be written (printing), and some tests that are already implicitly covered by other tests (but are not explicit yet.)

# TODO
- [ ] More minor tests
- [ ] 100% docstring coverage

# Open Questions
- This only operates on the C version of the type_description_interfaces struct... Do I need to write a C++ version? (Is there a way to convert between the two versions of the message, or are they otherwise interchangeable?)
  - Perhaps that C++ utils library would just have a conversion function, since we know the layout of the struct already???
- Should this be `type_description_interfaces_utils_c` instead? :grimacing:

### DIFF
--- a/type_description_interfaces_static/CMakeLists.txt
+++ b/type_description_interfaces_static/CMakeLists.txt
@@ -1,0 +1,46 @@
+cmake_minimum_required(VERSION 3.5)
+
+project(type_description_interfaces_static)
+
+# Default to C++14
+if(NOT CMAKE_CXX_STANDARD)
+  set(CMAKE_CXX_STANDARD 14)
+endif()
+if(CMAKE_COMPILER_IS_GNUCXX OR CMAKE_CXX_COMPILER_ID MATCHES "Clang")
+  add_compile_options(-Wall -Wextra -Wpedantic)
+endif()
+
+find_package(ament_cmake REQUIRED)
+find_package(rosidl_runtime_c REQUIRED)
+find_package(rosidl_runtime_cpp REQUIRED)
+
+file(GLOB sources "src/*.c")
+
+add_library(${PROJECT_NAME} ${sources})
+target_include_directories(${PROJECT_NAME} PUBLIC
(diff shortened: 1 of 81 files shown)


methylDragon:
Closing this in favor of https://github.com/ros2/rosidl/pull/728

PR #156, opened 2023-04-05
added a list of Parameter as a new message Parameters

saikishor:
Hello,

It would be interesting to have a list of Parameters as a new message type, so we can publish this info, rather than creating one per parameter

### DIFF
--- a/rcl_interfaces/CMakeLists.txt
+++ b/rcl_interfaces/CMakeLists.txt
@@ -24,6 +24,7 @@ rosidl_generate_interfaces(${PROJECT_NAME}
   "msg/ParameterEventDescriptors.msg"
   "msg/ParameterEvent.msg"
   "msg/Parameter.msg"
+  "msg/Parameters.msg"
   "msg/ParameterType.msg"
   "msg/ParameterValue.msg"
   "msg/SetParametersResult.msg"
--- a/rcl_interfaces/msg/Parameters.msg
+++ b/rcl_interfaces/msg/Parameters.msg
@@ -1,0 +1,4 @@
+# This is the message to communicate a list of parameters. It is a list of open 
+# structs with an enum in the descriptor to select which value is active.
+
+Parameter[] parameters 


saikishor:
Hello @gbiggs,

Pinging you on this.

clalancette:
This seems to overlap with the services that we have that provide largely the same thing: https://github.com/ros2/rcl_interfaces/tree/f03ee90089b96e9d6aa699019f43f17e1b0ac291/rcl_interfaces/srv

Also, parameters in ROS 2 are per-node, while the pub-sub bus is inherently anonymous.  So adding this doesn't seem like a good match to the way that parameters are implemented.

Can you describe more about your use case and what you are trying to accomplish with this?

saikishor:
Hello @clalancette!

Thanks for getting back to me. The idea to create this message is that there might be some use-cases where it is needed to broadcast the Parameters for other nodes to listen and configure them.  

> Can you describe more about your use case and what you are trying to accomplish with this?

For instance, the configuration of the robot could be a nice example. We have a robot with many configurations such as different arms, different lasers, different wheel types, etc. However, It is quite cumbersome to have to load the same file which is not encapsulated inside a ROS package again and again as this is basically the configuration of the robot. Here, we would like to latch a topic with the robot's configuration, so other nodes can listen and configure themself. 

I also thought it would make sense to make use of an existing standard message rather than creating a similar one for the use-case. It is also nice to have a standard way to retrieve info about the robot's configuration. It is something along the lines of how the `robot_description` now is being published along a latched topic.

What do you think?

Thank you,

Best Regards,
Sai Kishor Kothakota

saikishor:
Hello @clalancette,

What do you think about my previous response?

fujitatomoya:
@saikishor thanks for sharing the use case.

IMO,

> For instance, the configuration of the robot could be a nice example. We have a robot with many configurations such as different arms, different lasers, different wheel types, etc. However, It is quite cumbersome to have to load the same file which is not encapsulated inside a ROS package again and again as this is basically the configuration of the robot.

in that case, i would have the `robot_configuration_node` with all related parameters. so that other nodes can `read` the configuration data via parameter services instead of waiting the data is published. (this is one of the use case that i developed https://github.com/fujitatomoya/ros2_persist_parameter_server)

> Here, we would like to latch a topic with the robot's configuration, so other nodes can listen and configure themself.

i am not sure why we would want to go around like this. it would be much easier to call the service described above?

saikishor:
> in that case, i would have the robot_configuration_node with all related parameters. so that other nodes can read the configuration data via parameter services instead of waiting the data is published. (this is one of the use case that i developed https://github.com/fujitatomoya/ros2_persist_parameter_server)

Hello @fujitatomoya,

Thank you for the review and sharing the ROS2 persistent parameter server. It is definitely interesting and I've had a look into your package in the past. It's very interesting. Actually, we might use it too. Indeed we also have services to request parameters, but also this Publisher to get these bunch of variables all together.

I pretty much would like to stand with the ROS2 Standards, If this package of yours can be such a standard I'm happy to use. As I mentioned earlier, the `robot_description` is also being published right now, as the current framework doesn't allow you to have a global paramserver in ROS2. We would like to have something along the lines like  `tiago_robot_info`, where you can access all specific info of the robot. 

If we can make your package a standard for the ROS2 community, it would be really great. What do you think?

Thank you! :)

PR #159, opened 2023-08-24
Update the comments for SetParametersResult to reflect reality.

### LINKED ISSUE #2269: `add_on_set_parameters_callback` `reason` field is lost upon `successful=true`
## Bug report

**Required Info:**

- Operating System:
  - Jammy
- Installation type:
  - Binaries
- Version or commit hash:
  - Iron
- DDS implementation:
  - FastDDS
- Client library (if applicable):
  - rclcpp

#### Steps to reproduce issue

We are using the `add_on_set_parameters_callback` to set a UTM zone on a node.

```
  params_cb_handle_ = this->add_on_set_parameters_callback(
      std::bind(&GeoFrameNode::params_cb, this, std::placeholders::_1));
```

We had a test suite that was asserting on the `reason` field of the SetParameterResult msg and upon updating to Iron from Humble it no longer passes as `reason` is now empty:

```
        response = self.client_set_parameters.call(set_utm_request(ref_zone))
        assert response.results[0].successful is True
        assert response.results[0].reason == f"Reference UTM zone set to {ref_zone}."
```

#### Expected behavior

Not sure. If this is a deliberate change it probably should have been mentioned in the changelog and documented. If unintentional then it should probably be reverted to the behavior in humble unless there is a good reason.

clalancette:
One of the ways that parameter setting can fail is if a user callback (registered with add_on_set_parameters_callback()) rejects it.  These callbacks are chained in a Last-In-First-Called manner.

Another piece of information to be aware of is that the call to "set_parameters_atomically" has a single SetParametersResult return value.

Combining the above two pieces of information, then when "set_parameters_atomically" is called and one of the callbacks rejects it, it can easily set the reason string and return that to the user since no further callbacks are going to be called. However, if a callback accepts it but expects to set a reason code, one of the later callbacks can easily overwrite that return code.

Because of this, change the documentation to reflect the reality that a 'reason' can only be specified for the failure case. On success, 'reason' is undefined.

This should fix https://github.com/ros2/rclcpp/issues/2269

### DIFF
--- a/rcl_interfaces/msg/SetParametersResult.msg
+++ b/rcl_interfaces/msg/SetParametersResult.msg
@@ -2,6 +2,6 @@
 # successfully. A false value indicates the change was rejected.
 bool successful
 
-# Reason why the setting was either successful or a failure. This should only be
-# used for logging and user interfaces.
+# Reason why the setting was a failure. On success, the contents of this field
+# are undefined.  This should only be used for logging and user interfaces.
 string reason


fujitatomoya (approved):
lgtm

clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/19393/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/13922/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/20114/)

clalancette:
I'm going to go ahead and merge this one in; we can always change it again based on further feedback.

PR #161, opened 2023-09-05
Update the Log.msg constant types.

### LINKED ISSUE #158: rcl_interfaces Log message level field does not match type of constants
The rcl_interfaces Log message level field does not match the type of constants. The constants are of type byte and the field is of type uint8.

This means the following Python code fails with an exception:
```
from rcl_interfaces.msg import Log
msg = Log()
msg.level = Log.INFO
```

To get this to work, something like this needs to be done which is a bit ridiculous:
```
from rcl_interfaces.msg import Log
msg = Log()
msg.level = int.from_bytes(Log.INFO, byteorder='little')
```

Either the level field needs to be changed to a byte type or the constants should be changed to uint8. I don't think rclpy should be expected to do a silent conversion between types to make this work correctly.

clalancette:
These should be uint8 to match the type of the level field below, which is also uint8.  This will allow us to assign them without error when using Python.

This will fix #158

### DIFF
--- a/rcl_interfaces/msg/Log.msg
+++ b/rcl_interfaces/msg/Log.msg
@@ -12,21 +12,21 @@
 ##
 
 # Debug is for pedantic information, which is useful when debugging issues.
-byte DEBUG=10
+uint8 DEBUG=10
 
 # Info is the standard informational level and is used to report expected
 # information.
-byte INFO=20
+uint8 INFO=20
 
 # Warning is for information that may potentially cause issues or possibly unexpected
 # behavior.
-byte WARN=30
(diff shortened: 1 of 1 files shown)


clalancette:
CI:
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/19451/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/13974/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/20174/)

emersonknapp (approved):

PR #171, opened 2024-12-14
Added service_msgs dependency to solve a build error for a ROS 2 Jazzy workspace built from source

Mechazo11:
As part of a 3d robotics simulator, I put together a [ROS 2 Jazzy workspace](https://github.com/Mechazo11/ubuntu22_jazzy_ws) built from source files in Ubuntu 22.04.

When building all pertinent packages along with ```rcl_interfaces```, I ran into a dependency issue related to ```service_msgs```. This issue might have been reported last year in this answers.ros.org post: https://answers.ros.org/question/413630/issues-with-service_msgs/

I was able to solve this problem by simply adding the ```service_msgs``` dependency into `CmakeLists.txt` and `package.xml` files for the ```rcl_interfaces``` package and then tested to ensure it built correctly.

### DIFF
--- a/rcl_interfaces/CMakeLists.txt
+++ b/rcl_interfaces/CMakeLists.txt
@@ -14,6 +14,7 @@ endif()
 find_package(ament_cmake REQUIRED)
 find_package(rosidl_default_generators REQUIRED)
 find_package(builtin_interfaces REQUIRED)
+find_package(service_msgs REQUIRED)
 
 rosidl_generate_interfaces(${PROJECT_NAME}
   "msg/FloatingPointRange.msg"
--- a/rcl_interfaces/package.xml
+++ b/rcl_interfaces/package.xml
@@ -20,6 +20,7 @@
   <buildtool_depend>rosidl_default_generators</buildtool_depend>
 
   <depend>builtin_interfaces</depend>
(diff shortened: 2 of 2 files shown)


clalancette:
I'm not entirely opposed to this, but do you know why this is necessary?  We don't seem to need it on Ubuntu 24.04, RHEL-9, or Windows.  And there are no direct dependencies in this package to `service_msgs`.  Could you share the actual error you are getting when trying to compile?

Mechazo11:
Hi @clalancette sorry for the late response. I don't remember the exact error message. I will try to replicate the message again and send you a step-by-step instructions on how to replicating it in your end. Maybe I missed something cuz I agree, even I didn't understand why `service_msgs` was a dependency of `rcl_interface` even though without it, I couldn't get `rcl_interface` to build from source.

Mechazo11:
Hi @clalancette I am not sure what was going on with my initial attempts, I just cloned the latest version of rcl_interface and it seems to build just fine without the need for the `service_msgs` dependency. I will go ahead and close this PR for now. In future, if the error comes back I will open a ticket to discuss it further.

clalancette:
Thanks for looking into it!  And please do feel free to reopen if you see the problem again.

PR #172, opened 2025-03-27
Explicit Undeclareing parameter descriptor feild

William-Crosby:
Currently for node parameters to be undeclared they must also be dynamically typed. however I think that there are cases where statically typed params should be undeclared and dynamically typed params should be able to be removed. This adds another field in the parameter descriptor message that is by default false that represents whether or not a parameter is able to be undeclared.

### DIFF
--- a/rcl_interfaces/msg/ParameterDescriptor.msg
+++ b/rcl_interfaces/msg/ParameterDescriptor.msg
@@ -24,6 +24,9 @@ bool read_only false
 # If true, the parameter is allowed to change type.
 bool dynamic_typing false
 
+#if true, the parameter is allowed to be undeclared
+bool undeclarable false
+
 # If any of the following sequences are not empty, then the constraint inside of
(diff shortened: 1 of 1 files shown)


rcl_interfaces/msg/ParameterDescriptor.msg
```diff
@@ -24,6 +24,9 @@ bool read_only false
 # If true, the parameter is allowed to change type.
```
fujitatomoya: please see my comment on https://github.com/ros2/rclcpp/pull/2780#pullrequestreview-2736767882, and not inclined to take this PR as it offers, but this should be updated.

```suggestion
# If true, the parameter is allowed to change type and be undeclared.
```

fujitatomoya:
see https://github.com/ros2/rclcpp/pull/2780#issuecomment-2806972265 PMC conclusion for now.

PR #176, opened 2025-05-05
Add info to duration message and time message comments

### LINKED ISSUE #175: Duration.msg and Time.msg nanoseconds description
In both the duration message and time message definitions the `seconds` description is (roughly) as follows:

`Seconds component, range is valid over any possible int32 value.`

And the `nanoseconds` description is:

`Nanoseconds component in the range of [0, 1e9).`

So, the seconds description seems pretty explicit that it can be negative. My assumption is that this generally means that:

1. A time message can represent a time before the epoch
2. A duration message can represent a negative duration

Those make sense to me. 

But the nanoseconds description is weird to me. It says it is the "nanoseconds component". That makes sense for time and duration values with positive `seconds` values, but for a time or duration with a negative `seconds`, how does that work?

Specifically, how would a number between 0 and -1 be represented? Given that `nanoseconds` is of type `uint32`, it cannot be negative. But `seconds` is of type `int32`, and in C cannot represent negative zero. 

For example, take `-0.1 seconds`. How would it be represented? 

`seconds: 0, nanoseconds: -100000000` -- doesn't work because nanoseconds must be positive
`seconds: -0, nanoseconds: 100000000` -- doesn't work because negative zero isn't valid in C/isn't in the domain of int32. 
`seconds: -1, nanoseconds: 900000000` -- this would work, but goes against the description `Nanoseconds component in the range of [0, 1e9).`. It's also not very intuitive, but that's alright I guess if it works.

What is the expected way to represent this in C?

jimmy-mcelwain:
Closes #175

Makes comments for duration messages and time messages more clear, and adds examples for the message types.

### DIFF
--- a/builtin_interfaces/msg/Duration.msg
+++ b/builtin_interfaces/msg/Duration.msg
@@ -2,7 +2,7 @@
 # Messages of this datatype are of ROS Time following this design:
 # https://design.ros2.org/articles/clock_and_time.html
 
-# Seconds component, range is valid over any possible int32 value.
+# The seconds component, valid over all int32 values.
 int32 sec
 
 # The nanoseconds component, valid in the range [0, 1e9), to be added to the seconds component. 
--- a/builtin_interfaces/msg/Duration.msg
+++ b/builtin_interfaces/msg/Duration.msg
@@ -5,5 +5,8 @@
 # Seconds component, range is valid over any possible int32 value.
 int32 sec
 
-# Nanoseconds component in the range of [0, 1e9).
+# The nanoseconds component, valid in the range [0, 1e9), to be added to the seconds component. 
+# e.g.
+# The duration -1.7 seconds is represented as {sec: -2, nanosec: 3e8}
+# The duration 1.7 seconds is represented as {sec: 1, nanosec: 7e8}
 uint32 nanosec
--- a/builtin_interfaces/msg/Time.msg
+++ b/builtin_interfaces/msg/Time.msg
@@ -4,5 +4,8 @@
 # The seconds component, valid over all int32 values.
 int32 sec
 
-# The nanoseconds component, valid in the range [0, 1e9).
+# The nanoseconds component, valid in the range [0, 1e9), to be added to the seconds component. 
+# e.g.
+# The time -1.7 seconds is represented as {sec: -2, nanosec: 3e8}
+# The time 1.7 seconds is represented as {sec: 1, nanosec: 7e8}
 uint32 nanosec


builtin_interfaces/msg/Time.msg
```diff
@@ -4,5 +4,8 @@
 # The seconds component, valid over all int32 values.
 int32 sec
 
-# The nanoseconds component, valid in the range [0, 1e9).
+# The nanoseconds component, valid in the range [0, 1e9), to be added to the seconds component. 
+# e.g.
+# The time -1.7 seconds is represented as {sec: -2, nanosec: 3e8}
+# The time -.7 seconds is represented as {sec: 1, nanosec 7e8}
```
christophebedard:
```suggestion
# The time 1.7 seconds is represented as {sec: 1, nanosec: 7e8}
```

builtin_interfaces/msg/Duration.msg
```diff
@@ -2,8 +2,11 @@
...
 
-# Seconds component, range is valid over any possible int32 value.
+# The seconds component, valid over all int32 values.
 int32 sec
 
-# Nanoseconds component in the range of [0, 1e9).
+# The nanoseconds component, valid in the range [0, 1e9), to be added to the seconds component. 
+# e.g.
+# The duration -1.7 seconds is represented as {sec: -2, nanosec: 3e8}
+# The duration 1.7 seconds is represented as {sec: 1, nanosec 7e8}
```
christophebedard:
```suggestion
# The duration 1.7 seconds is represented as {sec: 1, nanosec: 7e8}
```

christophebedard (commented):
Thanks for the PR! I just have some minor comments.

tfoote (approved):
Thanks for the clarifications. LGTM with the suggestions from @christophebedard

fujitatomoya (approved):

ahcorde:
Pulls: ros2/rcl_interfaces#176
Gist: https://gist.githubusercontent.com/ahcorde/691446d19c1e39af9f13b1dc3f736add/raw/d09ed159a733c4a4f57d1586373f1227502ac9fb/ros2.repos
BUILD args:  --packages-above-and-dependencies rcl_interfaces
TEST args:  --packages-above rcl_interfaces
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/15918
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/23450/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/17632/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/2972/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/24278/)

christophebedard (approved):

ahcorde:
https://github.com/Mergifyio backport kilted jazzy humble

PR #185, opened 2025-10-15
Use builtin_interfaces/Time for TransitionEvent stamp

### LINKED ISSUE #184: Change `lifecycle_msgs/msg/TransitionEvent` to use `builtin_interfaces/msg/Time` instead of raw time stamp
### Description

Currently, the timestamp field of the `TransitionEvent` message is a uint64, which should be interpreted as a Unix timestamp.
However, most ROS interfaces use `builtin_interfaces/msg/Time` to represent the time.

### Motivation

This would make the timestamp of the transition event easier to work with, since it would work with the preexisting tools for other messages.

### Design / Implementation Considerations

The message would not change size (disregarding padding).

Currently this timestamp is not being filled by `rcl_lifecycle` (ros2/rcl#1019), so there is very little tooling directly affected.

### Additional Information

Is there a technical reason why this is a normal Unix timestamp instead of a `builtin_interfaces/msg/Time`?

I would be able to make the pull request.

EDIT: Extra additional information

SuperJappie08:
## Description

Use `builtin_interfaces/Time` for the timestamp in  `lifecycle_msgs/TransitionEvent`

Fixes #184 

Yes, the field name has been changed to be more in line with other uses (`stamp` instead of `timestamp`).
Furthermore, the type has been changed from `uint64` to `builtin_interfaces/Time` to allow it to be more easily used by preexisting tooling.

The impact will be minimal, since the `timestamp` field is currently not being filled in `rcl_lifecycle` (ros2/rcl#1019).

No.

### DIFF
--- a/lifecycle_msgs/msg/TransitionEvent.msg
+++ b/lifecycle_msgs/msg/TransitionEvent.msg
@@ -1,5 +1,5 @@
 # The time point at which this event occurred.
-uint64 timestamp
+builtin_interfaces/Time stamp
 
 # The id and label of this transition event.
 Transition transition
--- a/lifecycle_msgs/CMakeLists.txt
+++ b/lifecycle_msgs/CMakeLists.txt
@@ -13,6 +13,7 @@ endif()
 
 find_package(ament_cmake REQUIRED)
 find_package(rosidl_default_generators REQUIRED)
+find_package(builtin_interfaces REQUIRED)
 
 set(msg_files
   "msg/State.msg"
@@ -30,6 +31,7 @@ set(srv_files
 rosidl_generate_interfaces(${PROJECT_NAME}
   ${msg_files}
   ${srv_files}
+  DEPENDENCIES builtin_interfaces
   ADD_LINTER_TESTS
 )
 
--- a/lifecycle_msgs/package.xml
+++ b/lifecycle_msgs/package.xml
@@ -16,6 +16,9 @@
   <buildtool_depend>ament_cmake</buildtool_depend>
   <buildtool_depend>rosidl_default_generators</buildtool_depend>
 
+  <build_depend>builtin_interfaces</build_depend>
+
+  <exec_depend>builtin_interfaces</exec_depend>
   <exec_depend>rosidl_default_runtime</exec_depend>
 
   <test_depend>ament_lint_auto</test_depend>


fujitatomoya (approved):
as @SuperJappie08 mentions in issue header, this is a breaking change because it changes the topic message type. but until now, timestamp has never been used. that said, we can assume that there are no application that actually relies on this `timestamp` field.

i see that there could be almost no affection for the user application. but i would like to confirm with other PMC members on this.

wjwwood:
The only thing I would consider is whether or not we need a timestamp in this message at all? There is the sent and received timestamps for all messages, and usually we only include a timestamp in a message if it is representing data that may have happened in the past and there is an impactful delay in sending the message, or if we think the message will be relayed (and therefore have the sent timestamp overwritten by the second send).

lifecycle_msgs/msg/TransitionEvent.msg
```diff
@@ -1,5 +1,5 @@
 # The time point at which this event occurred.
-uint64 timestamp
+builtin_interfaces/Time stamp
```
wjwwood, line 2: If we do make this change, why change the field name? I would leave it as `timestamp`, as that is more descriptive anyway. I know `stamp` is used in other places (like in the std_msgs/Header), but that is mostly to match ROS 1.

tfoote:
Changing the type to be consistent makes sense to me. I think that this field has a legacy that predates builtin_interfaces/Time when we started with the uint64 nanoseconds representation more broadly. Especially since this field has not been being populated this seems like the right time to update the type and then fill in the implementation with the new type. 

I agree with William though that I'd bias towards keeping the name so that it's a smaller change. We have the shorthand in Header but I think that the full variable name is valuable without the header context. stamp is somewhat overloaded for a standalone field.

fujitatomoya:
@wjwwood @tfoote thanks for the comments.

i now also think having timestamp field is redundant, at least i do not find it useful for the application.

@SuperJappie08 what do you think? probably we could just remove this field instead?

lifecycle_msgs/msg/TransitionEvent.msg
SuperJappie08: I also changed the name to be consistent with other messages, such as `rcl_interfaces/msg/Log` and others.

Isn't there also something to say in terms of convention here?
Because I could imagine it being 'weird' (for lack of a better term) if this is the only message, which timestamp is not named `stamp`

(I'm more than happy to change it if that makes more sense.)

gavanderhoorn:
> The only thing I would consider is whether or not we need a timestamp in this message at all?

Might be interesting to compare this to some sw-defined fieldbuses, such as OPC-UA.

In those systems, the approach is often to have different stamps for different 'aspects' of what happened to a message. So a timestamp for when an event occurred, one for when the message was sent (which is not necessarily always when the event happened, nor would it necessarily be sent by the source of the event) and sometimes even a stamp to record when it was received (which again might not always be when the consumer gets its hands on it).

With that in mind I could see a use for this additional `timestamp` field here.

SuperJappie08:
> i now also think having timestamp field is redundant, at least i do not find it useful for the application.
> 
> @SuperJappie08 what do you think? probably we could just remove this field instead?

An application I have/had is to get the exact moment of a transition as a time reference for a measurement that is controlled by a lifecycle node.
I need exact timing since it is embedded hardware and control related, but since it isn't a real-time system, I'm dependent on accurate timestamps.

lifecycle_msgs/msg/TransitionEvent.msg
wjwwood: Yeah, I don't feel strongly about it. `stamp` is fine, even if I think `timestamp` is a better name in a vacuum.

wjwwood:
Those sound like reasonable use cases for a timestamp in the message (since there's no plan at this time to have a "created" timestamp built-in to the middleware), so leaving it is fine. I just wanted to check if it was needed (since it seems to have gone so long being unused).

Also, as I said I'm not too worried about the field name. `stamp` is more consistent with other parts of the stack, but changing as little as possible is best in general when changing messages. Either should be ok.

fujitatomoya:
> a timestamp for when an event occurred, one for when the message was sent (which is not necessarily always when the event happened, nor would it necessarily be sent by the source of the event) and sometimes even a stamp to record when it was received (which again might not always be when the consumer gets its hands on it).

i think this makes sense to me. i was thinking that `rmw_message_info` as following can cover this requirement, that is the major reason i do not find the use of this timestamp field in this message.

```console
...<snip>
publication_sequence_number: 4
publisher_gid:
  data: !!binary |
    AQ8E9kv3XYQAAAAAAAATAw==
  implementation_identifier: rmw_fastrtps_cpp
received_timestamp: 1761111785570276567
reception_sequence_number: null
source_timestamp: 1761111785570084585
---
timestamp: 0
transition:
  id: 0
  label: ''
start_state:
  id: 13
  label: activating
goal_state:
  id: 3
  label: active
---
...<snip>
```

but as @gavanderhoorn mentioned, source and receive timestamp is different when the event takes place.

and @SuperJappie08 thanks for iterating with us. i will go ahead to run CI.

fujitatomoya:
Pulls: ros2/rcl_interfaces#185, ros2/rcl#1269, ros2/rclcpp#2967, ros2/rclpy#1528
Gist: https://gist.githubusercontent.com/fujitatomoya/43f97e6f8303a55ef1fdd1c645c51dd1/raw/feab0bba51e7c64b9f07536c61c03781c6dcd1e2/ros2.repos
BUILD args:  --packages-above-and-dependencies lifecycle_msgs rcl_lifecycle rclcpp_lifecycle rclpy
TEST args:  --packages-above lifecycle_msgs rcl_lifecycle rclcpp_lifecycle rclpy
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/17335
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/25484/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/19561/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/5012/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/25775/)

fujitatomoya:
Pulls: ros2/rcl_interfaces#185, ros2/rcl#1269, ros2/rclcpp#2967, ros2/rclpy#1528
Gist: https://gist.githubusercontent.com/fujitatomoya/7419b9893cf173e26ca59097d1d673aa/raw/feab0bba51e7c64b9f07536c61c03781c6dcd1e2/ros2.repos
BUILD args:  --packages-above-and-dependencies lifecycle_msgs rcl_lifecycle rclcpp_lifecycle rclpy
TEST args:  --packages-above lifecycle_msgs rcl_lifecycle rclcpp_lifecycle rclpy
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/17401
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/25608/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/19641/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/5169/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/25842/)

SuperJappie08:
@fujitatomoya, there was one issue in `launch_ros`, which should be fixed by ros2/launch_ros#495

fujitatomoya:
@SuperJappie08 there is a couple of compiler and clang warning, https://ci.ros2.org/job/ci_linux-rhel/5169/. can you also check them?

SuperJappie08:
> @SuperJappie08 there is a couple of compiler and clang warning, https://ci.ros2.org/job/ci_linux-rhel/5169/. can you also check them?

@fujitatomoya, I left some comments on the PR on ros2/rcl#1269 ([comment](https://github.com/ros2/rcl/pull/1269#issuecomment-3495882667)).
If the comments don't require action, could you retry the CI, because it should be fixed (when ros2/launch_ros#495 is included).

fujitatomoya:
Pulls: ros2/rcl_interfaces#185, ros2/rcl#1269, ros2/rclcpp#2967, ros2/rclpy#1528, ros2/launch_ros#495
Gist: https://gist.githubusercontent.com/fujitatomoya/509dc37f463751da80d68854d37ec1d0/raw/5f7a4fa0c2b57380aa59f72196e8279872b6a096/ros2.repos
BUILD args:  --packages-above-and-dependencies lifecycle_msgs rcl_lifecycle rclcpp_lifecycle rclpy launch_ros
TEST args:  --packages-above lifecycle_msgs rcl_lifecycle rclcpp_lifecycle rclpy launch_ros
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/18088
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/27397/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/20604/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/7320/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/26559/)

fujitatomoya:
they all passed (eventually with retries)

For completeness:

- Linux [Build Status](https://ci.ros2.org/job/ci_linux/27416/)
- Linux-aarch64 [Build Status](https://ci.ros2.org/job/ci_linux-aarch64/20604/)
- Linux-rhel [Build Status](https://ci.ros2.org/job/ci_linux-rhel/7325/)
- Windows [Build Status](https://ci.ros2.org/job/ci_windows/26559/)

fujitatomoya:
@SuperJappie08 thanks for being patient and making contribution!

PR #186, opened 2026-02-20
Add 'executor_type' and 'thread_num' to LoadNode

armaho:
## Description

This pull request is related to ros2/rclcpp#2930 and ros2/rclcpp#3080. It adds the parameters `executor_type` and `thread_num` to `composition_interfaces/srv/LoadNode`.

No.

### Additional Information

Nothing.

### DIFF
--- a/composition_interfaces/srv/LoadNode.srv
+++ b/composition_interfaces/srv/LoadNode.srv
@@ -28,6 +28,19 @@ rcl_interfaces/Parameter[] parameters
 
 # key/value arguments that are specific to a type of container process.
 rcl_interfaces/Parameter[] extra_arguments
+
+# Executor used for this node. Leave empty to share the executor
+# used for the ComponentManager. Otherwise, a new executor will
+# be used for this node. Options:
+#   + SingleThreadedExecutor
+#   + MultiThreadedExecutor
+string executor_type
+
+# Number of threads dedicated to the executor for this node. This option
+# is only valid if you choose MultiThreadedExecutor as the executor type.
+# Leave empty to use the default.
+uint8 thread_num
+
 ---
 # True if the node was successfully loaded.
 bool success


composition_interfaces/srv/LoadNode.srv
```diff
@@ -28,6 +28,19 @@ rcl_interfaces/Parameter[] parameters
 
 # key/value arguments that are specific to a type of container process.
 rcl_interfaces/Parameter[] extra_arguments
+
+# Executor used for this node. Leave empty to share the executor
+# used for the ComponentManager. Otherwise, a new executor will
+# be used for this node. Options:
+#   + SingleThreadedExecutor
+#   + MultiThreadedExecutor
+string executor_type
```
fujitatomoya, line 37: i think this makes more sense to keep this as arbitrary string type if user can configure their own defined executor to the component manager. otherwise, maybe predefined enum would be enough? can we not use `EventsExecutor` here?

composition_interfaces/srv/LoadNode.srv
```diff
@@ -28,6 +28,19 @@ rcl_interfaces/Parameter[] parameters
...
+
+# Number of threads dedicated to the executor for this node. This option
+# is only valid if you choose MultiThreadedExecutor as the executor type.
+# Leave empty to use the default.
+uint8 thread_num
```
fujitatomoya, line 42: what is the default thread number for MultiThreadedExecutor? hardware thread num?

armaho: yes. I will specify that.

composition_interfaces/srv/LoadNode.srv
armaho: > i think this makes more sense to keep this as arbitrary string type if user can configure their own defined executor to the component manager. 

User cannot do that. This is the executor used for the node itself, so `ComponentContainer` itself must be able to construct an object of the specified executor.

> otherwise, maybe predefined enum would be enough?

Yeah, sure. I will do that.

> can we not use EventsExecutor here?

There's nothing stoping us. I will add that.

mjcarroll:
I don't think that this is the right thing to do. I would be on-board with command line or ROS parameter interfaces for configuring the executor for a component container.  

Having multiple (heterogenous) kinds of executors in a single container starts to feel like it may be an application of _manual compile-time composition_ rather than needing the flexibility of runtime dynamic composition.

Curious what is driving this need.

fujitatomoya:
@armaho can we close this too?

armaho:
@fujitatomoya 

The original issue is resolved, and we can close this. However, you also wanted the ability to specify the type of executor used inside `ComponentContainer` for each node separately. This is not addressed by #3134. If it's still needed, I can work on it via a new issue.

I'm going to close this PR for now.

fujitatomoya:
> you also wanted the ability to specify the type of executor used inside ComponentContainer for each node separately. This is not addressed by #3134. If it's still needed, I can work on it via a new issue.

@armaho i think this is good feature for the component container, what do you think? i suggest that we can create the issue 1st and then get feedback from what other people think. how about that?

armaho:
@fujitatomoya 

A new issue would be a good start, cause I think that @mjcarroll has also had some comments about this. I’m still new here, so I don’t know enough about the community’s needs to comment on this but I also think it's good to be able to have different types of executor within a container.

PR #188, opened 2026-03-18
Add Graph description messages to rosgraph_msgs

### LINKED ISSUE #182: "ROS Graph" messages to describe an entire ROS application
### Description

Proposal for discussion and consideration:

Add here to the core a set of messages that can describe a running ROS graph in full (or as full as possible).
This is approximately:

```
# Graph
Node[] nodes

# Node
string fully_qualified_name
rcl_interfaces/ParameterDescriptor[] parameters
rcl_interfaces/ParameterValue[] parameter_values
Topic[] publishers
Topic[] subscriptions
Service[] clients
Service[] servers
Action[] action_clients
Action[] action_servers

# Topic
string name
string type
QosProfile qos

# etc...
```

### Motivation

Having a common data representation for the ROS Graph is a key part of toplevel health monitoring, which allows us and run analyses on an individual graph for connectivity and compatibility, etc, or to set an "expected observed graph" and diff running or recorded graphs against those expectations.

### Design / Implementation Considerations

Consider all the aspects that entities within the graph can have. We have not yet actually considered services or actions, but they shouldn't be much extra information.

Type hashes would be a valuable addition to topics, starting in Jazzy, though I want to add these messages to all live distros (meaning Humble, too). Possible the full Type Description would be desired.

### Additional Information

We currently have this implemented at https://github.com/ros-tooling/graph-monitor/tree/main/rosgraph_monitor_msgs/msg, but think it makes a lot of sense as a core concept, that the graph monitoring library can produce.

emersonknapp:
## Description

Fixes #182 

Add new message definitions to `rosgraph_msgs` that can be used to describe a running or potential ROS Graph.

- `Graph`: a list of nodes
- `Node` : describe a single node's parameters and interface endpoints
- `Topic`: describe a single topic endpoint, either Publisher or Subscription
- `Service`: describe a single service endpoint, either Server or Client
- `Action`: describe a single action endpoint, either Server or Client
- `QoSProfile`: describe a QoS Profile, either offered or requested
-  `InterfaceType`: identify a `rosidl` type
- `TypeHash`: identify a REP-2016 RIHS type hash (`rosidl_runtime_c` defined type)

This is intended to supplant the definitions in use at https://github.com/ros-tooling/graph-monitor/tree/main/rosgraph_monitor_msgs

Yes

No

### Additional Information

Part of the [NoDL](https://github.com/ros-tooling/nodl) and [rosgraph_monitor](https://github.com/ros-tooling/graph-monitor) work being done within the ROSGraph Working Group.

We would like to get these messages into the core for Lyrical, so that we can build tools around them for release into that LTS distro.

These messages act as a counterpart to the NoDL schema, being used to output either a desired or observed state of a ROS application as a message between nodes. The NoDL schema is not a strict duplicate of the information that these messages will contain, because:
1. NoDL will contain information relevant to code generation that is of more detail than can be observed by graph inspection
2. These messages contain fields about "instantiated values" that do not exist at specification time, such as actual parameter values.
3. The contents of these messages generally refer to resolved fully-qualified graph names after namespacing and remapping are applied, something that is not known or relevant at the node specification level.


wjwwood (approved):
Looks reasonable to me

skyegalaxy (approved):
seems like pretty much exactly what was discussed at this week's ROSgraph WG

rosgraph_msgs/msg/QoSProfile.msg
```diff
@@ -0,0 +1,42 @@
...
+uint8 HISTORY_KEEP_LAST=1
+uint8 HISTORY_KEEP_ALL=2
+uint8 HISTORY_UNKNOWN=3
+uint8 history
+
+# Reliability policy
+uint8 RELIABILITY_SYSTEM_DEFAULT=0
+uint8 RELIABILITY_RELIABLE=1
+uint8 RELIABILITY_BEST_EFFORT=2
+uint8 RELIABILITY_UNKNOWN=3
```
fujitatomoya, line 27: we do not need `best_available`? actual QoS configuration will be set in other available policies above at runtime, but user can set this `best_effort` QoS in NoDL yaml. that means, it needs to be able to verify this `best_effort` setting against the what https://github.com/ros-tooling/graph-monitor/tree/main/rosgraph_monitor collects at runtime? I am not sure how we can verify this situation (i believe that node graph collected by rmw uses actual QoS), because actual QoS will not be `best_effort` once node starts running...

aside from above, i think it would be better to add `RELIABILITY_BEST_AVAILABLE` here to keep the compatibility with rmw interface. so that we will not miss-align the next policy.

```suggestion
uint8 RELIABILITY_UNKNOWN=3
uint8 RELIABILITY_BEST_AVAILABLE=4
```

rosgraph_msgs/msg/QoSProfile.msg
```diff
@@ -0,0 +1,42 @@
...
+uint8 reliability
+
+# Durability policy
+uint8 DURABILITY_SYSTEM_DEFAULT=0
+uint8 DURABILITY_TRANSIENT_LOCAL=1
+uint8 DURABILITY_VOLATILE=2
+uint8 DURABILITY_UNKNOWN=3
```
fujitatomoya, line 35: same here.

```suggestion
uint8 DURABILITY_UNKNOWN=3
uint8 DURABILITY_BEST_AVAILABLE=4
```

rosgraph_msgs/msg/QoSProfile.msg
```diff
@@ -0,0 +1,42 @@
...
+uint8 durability
+
+# Liveliness policy
+uint8 LIVELINESS_SYSTEM_DEFAULT=0
+uint8 LIVELINESS_AUTOMATIC=1
+uint8 LIVELINESS_MANUAL_BY_TOPIC=3
+uint8 LIVELINESS_UNKNOWN=4
```
fujitatomoya, line 43:
```suggestion
uint8 LIVELINESS_UNKNOWN=4
uint8 LIVELINESS_BEST_AVAILABLE=5
```

rosgraph_msgs/msg/Service.msg
```diff
@@ -0,0 +1,6 @@
+# Describes a single Service endpoint, which may be a Server or Client
```
fujitatomoya, line 1: why this does not have QoSProfile?

rosgraph_msgs/msg/Action.msg
```diff
@@ -0,0 +1,8 @@
+# Describes a single Action endpoint, which may be a Server or Client
```
fujitatomoya, line 1: why this does not have QoSProfile?

rosgraph_msgs/msg/Service.msg
emersonknapp: You're right - for some reason I thought QoS was hidden under the hood for Services and Actions, but they are exposed. Just not commonly set to anything other than the default. I will add

rosgraph_msgs/msg/QoSProfile.msg
emersonknapp: Good catch! I wrote the first draft of this proposal against Humble which didn't have that value yet. I did not intentionally leave it out. I'll make sure it matches the Rolling values now.

skyegalaxy (changes requested):
need to add DCO for the last commit

emersonknapp:
Note for reviewers - I'm updating the docstrings with more information, and calling out the change here.

I was trying to use these messages to represent both the "abstract descriptor" of these entities, as well as the "observed/observable instantiation" of them. However, those two concepts have some key disjunctions.

Therefore, I am focusing in on the latter - observed/observable instantiation. We will use NoDL for the "abstract descriptor" and may or may not choose to be able to represent it as messages, which will be a separate exercise.

Thanks @fujitatomoya for taking the close look that forced this decision

rosgraph_msgs/msg/Service.msg
emersonknapp: I had to reconcile the fact that a service is created with a single QoS input, but is observed with two actual response/request QoS results. 

I have scoped this definition to be "actual" / "observed" state, which means 2 QoS

rosgraph_msgs/msg/Action.msg
emersonknapp: I had to think really hard about this one, too. The user is given the option to specify 5 QoS profiles as input - 3 services and 2 topics. This resolves to 8 QoS profiles observable at runtime (2 per service). I've gone that direction and defined an action as only an aggregate type of 3 services + 2 topics + a name.

rosgraph_msgs/msg/TypeHash.msg
emersonknapp, line 1: I wonder if instead of a structure here we should just allow this to be the stringified representation of the type hash that we show in the ros2 cli, etc. Theoretically a RIHS v2 wouldn't have to have the same length, so the `[32]` fixed size below wouldn't necessarily provide the flexibility to communicate that.

(hightly doubt we ever define RIHS02, but hey, we allowed for it explicitly)

Could just do non-fixed-size array, which would be half the size of sending stringified hexadecimal.

rosgraph_msgs/msg/Node.msg
```diff
@@ -0,0 +1,24 @@
...
+
+# Communications endpoints - Topics, Services, and Actions
+Topic[] publishers
+Topic[] subscriptions
+
+Service[] service_clients
+Service[] service_servers
+
+Action[] action_clients
+Action[] action_servers
```
skyegalaxy, line 24: Would these lists allow for duplicates, say for example if a node spins up multiple (action) clients to the same (action) server for whatever reason?

rosgraph_msgs/msg/TypeHash.msg
skyegalaxy: I personally would err on the side of the fixed size array for the type hash and just increase the size when we actually need to, and feel less inclined to allow for a variable length string. Part of that is [because of the known issues with rosidl / zero copy and variable length data](https://discourse.openrobotics.org/t/is-it-time-to-discuss-rosidl), but maybe zero-copy isn't going to matter anyway if there are other `string` fields inside of the message. It could also be the case someday that as a result of these discussions, strings in ROS messages will end up with a max character length before being dropped as the implicit upper bound of a RIHSv2 length.

skyegalaxy (approved):

rosgraph_msgs/msg/TypeHash.msg
emersonknapp: there are many string fields in here :sweat_smile:

rosgraph_msgs/msg/Node.msg
emersonknapp: nothing prevents it. based on the current representation you wouldn't be able to tell them apart though.

I'm now wondering if I need to basically have the leaf message actually be an encoding of [`rmw_topic_endpoint_info`](https://github.com/ros2/rmw/blob/rolling/rmw/include/rmw/topic_endpoint_info.h#L31) (and service_endpoint_info), and include the GID as well.

That would directly represent the "observable state" as it's defined most fundamentally in the API, if we try to match those definitions.

fujitatomoya (approved):

christophebedard:
@emersonknapp is this ready to merge? I can run CI

christophebedard:
Pulls: ros2/rcl_interfaces#188
Gist: https://gist.githubusercontent.com/christophebedard/854094017ffcb7dbc8374893da5aeb99/raw/7a802edfc8d7a6d4ce2b7610570e86bd9e86bf9e/ros2.repos
BUILD args:  --packages-above-and-dependencies rosgraph_msgs
TEST args:  --packages-above rosgraph_msgs
ROS Distro: rolling
Job: ci_launcher
ci_launcher ran: https://ci.ros2.org/job/ci_launcher/18808
* Linux [Build Status](http://ci.ros2.org/job/ci_linux/28500/)
* Linux-aarch64 [Build Status](http://ci.ros2.org/job/ci_linux-aarch64/21364/)
* Linux-rhel [Build Status](http://ci.ros2.org/job/ci_linux-rhel/8356/)
* Windows [Build Status](http://ci.ros2.org/job/ci_windows/27366/)

skyegalaxy:
- Linux-rhel [Build Status](https://ci.ros2.org/job/ci_linux-rhel/8415/)
- Windows [Build Status](https://ci.ros2.org/job/ci_windows/27427/)

emersonknapp:
CI failures unrelated, merging

emersonknapp:
@mergifyio backport humble jazzy kilted

